PR #5, opened 2019-11-24
chore: create workflow file

noahnu:


.github/workflows/pythonapp.yml
```diff
@@ -0,0 +1,35 @@
...
+        python -m pip install -e . -r requirements.txt -r dev-requirements.txt
+    - name: Lint
+      run: |
+        invoke lint
+    - name: Test
+      run: |
+        invoke test
+    - name: Build Distribution
+      run: |
+        python setup.py sdist bdist_wheel
```
iamogbz: Could probably make an invoke command

iamogbz: Should also stub out a release step here
```suggestion
        inv build
    - name: Publish
      run: |
        inv publish
```

script/bootstrap
```diff
@@ -20,6 +20,6 @@ fi
 VENV="${PWD##*/}.venv"
 test -d $VENV || python3 -m venv $VENV || return
 $VENV/bin/pip install -U pip --no-cache-dir
-$VENV/bin/pip install -e . -r requirements.txt
+$VENV/bin/pip install -e . -r requirements.txt -r dev-requirements.txt
```
iamogbz: Since the dev-requirements.in already include requirements.in
```suggestion
$VENV/bin/pip install -e . -r dev-requirements.txt
```

.github/workflows/pythonapp.yml
noahnu: I'll do this in a follow up

PR #7, opened 2019-11-24
ci: stub out release command

iamogbz:
# Description

Only releases on the python `3.7` build
All builds get published to `test.pypi.org`
Valid version tagged builds. e.g. `v0.1.0` etc. get published to`pypi.org` and test

## DevQA

- `inv release` should dry run
- `inv release --no-dry-run` should fail with message saying run in CI
- `env CI=1 && inv release --no-dry-run` should only publish version to test.pypi.org
- `git tag v0.1.0`
- `env CI=1 && inv release --no-dry-run` should now publish version to both test.pypi.org and pypi.org


tasks.py
```diff
@@ -1,18 +1,64 @@
...
+
+
+@task
+def build(ctx):
+    ctx.run("python setup.py sdist bdist_wheel")
+
+
+@task
+def publish(ctx, dry_run=True):
+    print(
```
iamogbz: After environment variables `TWINE_USERNAME` and `TWINE_PASSWORD` is setup
```suggestion
    ctx.run(
```

PR #9, opened 2019-11-24
chore: type check with mypy in lint task

iamogbz:
# Description

Slight code changes needs to make it work

Also adds precommit linting

## DevQA

- `inv lint`
- `inv lint --fix`


iamogbz:
@all-contributors please add @iamogbz for code, design and infra

PR #14, opened 2019-12-05
ci: enable deploy

iamogbz:
Enables release on tagged builds

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -1,6 +1,6 @@
 name: Syrupy CICD
 
-on: [push, pull_request]
+on: [push]
 
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -32,5 +32,8 @@ jobs:
       if: matrix.python-version != '3.7'
       run: invoke build
     - name: Release Python 3.7
+      env:
+        TWINE_USERNAME: ${{ secrets.TWINE_USERNAME }}
```
noahnu: Is this stripped out of logs by github?

iamogbz: Yup yup, you can check the output

PR #18, opened 2019-12-05
fix: assert match should raise assertion

### LINKED ISSUE #17: assert_match does not fail assertion
**Describe the bug**

`snapshot.assert_match(ANY)` always passed

**To Reproduce**
1. Replace with `snapshot.assert_match(None)`
https://github.com/tophat/syrupy/blob/a5c46b15942e4d75c9470b24f3e2bfc9fa661087/tests/test_snapshots.py#L14
2. Run command `inv test`
3. See that test does not fail

**Expected behavior**

Should fail with the same error as `assert None == snapshot`

iamogbz:
# Description

Fixes #17

## DevQA

Follow issue STR

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -62,8 +62,8 @@ class SnapshotAssertion:
             session=self._session,
         )
 
-    def assert_match(self, data) -> bool:
-        return self._assert(data)
+    def assert_match(self, data):
+        assert self == data
 
     def get_assert_diff(self, data) -> List[str]:
         deserialized = self._recall_data(index=self.num_executions - 1)
--- a/tests/test_snapshots.py
+++ b/tests/test_snapshots.py
@@ -11,7 +11,7 @@ def test_raw_string(snapshot):
 
 def test_multiple_snapshots(snapshot):
     assert "First." == snapshot
-    assert "Second." == snapshot
+    snapshot.assert_match("Second.")
     assert "Third." == snapshot
 
 


src/syrupy/assertion.py
```diff
@@ -62,8 +62,8 @@ def with_class(
             session=self._session,
         )
 
-    def assert_match(self, data) -> bool:
-        return self._assert(data)
+    def assert_match(self, data):
+        assert self == data
```
noahnu: Does pytest report the failure in syrupy/assertion.py though or in the original test?

iamogbz: The failure is in the `assertion.py` file but the full trace shows the tests it fails in.
```sh
============================== FAILURES ===============================
_______________________ test_multiple_snapshots _______________________

snapshot = <SnapshotAssertion (2)>

    def test_multiple_snapshots(snapshot):
        assert "First." == snapshot
>       snapshot.assert_match("Second")

tests/test_snapshots.py:14: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <SnapshotAssertion (2)>, data = 'Second'

    def assert_match(self, data):
>       assert self == data
E       assert snapshot == Second
E         - Second
E         + Second.

src/syrupy/assertion.py:66: AssertionError
===================== 1 failed, 8 passed in 0.28s =====================
```

iamogbz: Same behaviour with snapshottest at the moment

PR #25, opened 2019-12-06
feat: show more shapshot assertion information

### LINKED ISSUE #12: Show number of snapshots asserted in the report

Show how many snapshots were used in the test session

Showing the number summary e.g.

```txt
1 snapshot failed | 5 snapshots passed | 1 snapshot updated | 3 snapshots generated | 2 snapshots unused
```

N/A

**Additional context**

This requires an overhaul of how snapshot assertions are tracked

iamogbz:
## Description

Closes #12 

Show summary of snapshots
- [x] Failed - existed, assert failed
- [x] Passed - existed, no changes
- [x] Updated - existed, with changes
- [x] Missing/Written - did not exist, assert failed/succeeds
- [x] Unused - existed, not asserted

### QA
- `inv test -d`
- Profit!


src/syrupy/assertion.py
```diff
@@ -10,12 +13,20 @@
...
-    from .serializers.base import AbstractSnapshotSerializer
     from .location import TestLocation
+    from .serializers.base import AbstractSnapshotSerializer
     from .session import SnapshotSession
+    from .types import SnapshotFiles
+
+
+AssertionResult = namedtuple(
+    "AssertionResult",
+    ["asserted", "created", "file", "name", "recalled", "success", "updated"],
```
noahnu: Preference for attrs: http://www.attrs.org/en/stable/why.html#namedtuples

Allows us to be stricter with types.

src/syrupy/assertion.py
```diff
@@ -31,6 +42,7 @@ def __init__(
         self._serializer_class = serializer_class
         self._test_location = test_location
         self._executions = 0
+        self._execution_results: Dict[int, AssertionResult] = {}
```
noahnu: Perhaps we should stick to forward ref style, i.e. quotes, for all class types? For consistency

src/syrupy/assertion.py
```diff
@@ -39,14 +51,25 @@ def __init__(
...
     def num_executions(self) -> int:
         return int(self._executions)
 
+    @property
+    def executions(self) -> Dict[int, AssertionResult]:
+        return self._execution_results
+
+    @property
+    def discovered_snapshots(self) -> "SnapshotFiles":
+        return {
```
noahnu: What's the reasoning for moving this into the assertion class?

src/syrupy/serializers/base.py
```diff
@@ -104,7 +101,7 @@ def write(self, data: "SerializableData", index: int = 0) -> None:
         self.write_snapshot_or_remove_file(snapshot_file, snapshot_name, data)
 
     def post_write(self, data: "SerializableData", index: int = 0) -> None:
-        self._snap_file_hook(index)
+        pass
```
noahnu: Can we think of a use case for the hooks? I.e. an argument for keeping them?

src/syrupy/session.py
```diff
@@ -181,16 +197,3 @@ def _diff_snapshot_files(
 
     def _count_snapshots(self, snapshot_files: "SnapshotFiles") -> int:
         return sum(len(snaps) for snaps in snapshot_files.values())
-
-    def _in_snapshot_dir(self, path: str) -> bool:
-        parts = path.split(os.path.sep)
-        return SNAPSHOT_DIRNAME in parts
-
-    @lru_cache(maxsize=32)
-    def _walk_dir(self, root: str) -> Generator[str, None, None]:
```
noahnu: I know this is not new, but how does an LRU cache work with a generator here? Can you give a short example of a case where the cache provided values? Having trouble visualizing it for walk dir.

src/syrupy/assertion.py
iamogbz: It's pretty much the same logic, but now each assertion has a reference to the snapshots it discovers

src/syrupy/serializers/base.py
iamogbz: We don't use it anymore since each assertion keeps track of it's snapshots, but they could be useful for custom plugins performing actions at different stages of serializing

src/syrupy/session.py
iamogbz: I removed it in this PR. But an example would be we call this multiple times when discovering snapshots in the same test file using different assertion fixtures, the general idea is we would not need to walk the entire `__snapshot__` directory each time if we memoize it.

src/syrupy/assertion.py
iamogbz: Missed it

src/syrupy/assertion.py
iamogbz: mypy does not seem to play nice with `attr.make_class`

src/syrupy/utils.py
```diff
@@ -0,0 +1,19 @@
+import os
+from functools import lru_cache
```
iamogbz:
```suggestion
```

src/syrupy/assertion.py
noahnu: What doesn't work? Can't you just use the attr.s decorator?

src/syrupy/serializers/base.py
noahnu: We can leave it for now but I think it's better to add it only when we find a use case. Rather than add functionality that won't be used because we're trying to guess at how the community will use this.

src/syrupy/assertion.py
iamogbz: Yeah but it's more verbose 😅
```py
AssertionResult = attr.make_class(
```
vs
```py
@attr.s
class AssertionResult(object):
    asserted = attr.ib()
    created = attr.ib()
    ...
```

src/syrupy/session.py
iamogbz: Discussed offline `lru_cache` does not support generators, removed in this PR

PR #26, opened 2019-12-06
fix: compare snapshots without deserializing

### LINKED ISSUE #21: Compare snapshots without deserializing

Snapshots are currently deserialized into python objects before comparison is done, this is unnecessary

No deserialization is done and comparison is done on the serialized representation

N/A

**Additional context**

Unlocks using `yaml.dumps` in place of `yaml.safe_dumps` because loading yaml data is needless

iamogbz:
# Description

Fixes #21 

- [x] Do not deserialize to assert
- [x] Use `yaml.dump`
- [x] Do not rewrite matching snapshots

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -91,15 +91,15 @@ class SnapshotAssertion:
     def _assert(self, data) -> bool:
         self._session.register_assertion(self)
         try:
+            serialized_data = self.serializer.encode(data)
             if self._update_snapshots:
-                serialized_data = self.serializer.encode(data)
                 self.io.create_or_update_snapshot(
                     serialized_data, index=self.num_executions
                 )
                 return True
 
-            deserialized = self._recall_data(index=self.num_executions)
-            if deserialized is None or data != deserialized:
+            snapshot_data = self._recall_data(index=self.num_executions)
+            if snapshot_data is None or serialized_data != snapshot_data:
                 return False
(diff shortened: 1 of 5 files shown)


src/syrupy/io.py
```diff
@@ -128,7 +128,7 @@ def _read_file(self, filepath: str) -> Any:
         """
         try:
             with open(filepath, "r") as f:
-                return yaml.safe_load(f) or {}
+                return yaml.load(f, Loader=yaml.FullLoader) or {}
```
iamogbz: Still needed to parse the document into snapshot names

src/syrupy/assertion.py
```diff
@@ -90,23 +91,22 @@ def __eq__(self, other) -> bool:
...
         self._session.register_assertion(self)
         try:
-            if self._update_snapshots:
-                serialized_data = self.serializer.encode(data)
+            serialized_data = self.serializer.encode(data)
+            snapshot_data = self._recall_data(index=self.num_executions)
+            matches = snapshot_data is not None and serialized_data == snapshot_data
+            if matches:
+                return True
+            if not matches and self._update_snapshots:
```
noahnu: (nb) if you move this condition before line 97, you can then just return `matches`. I.e. 2 if conditions instead of 3.

iamogbz: Will do this

PR #30, opened 2019-12-07
refactor: collapse io and serializer into single class

noahnu:
So far we haven't found a use case for separating IO and Serializer logic. I think we can simplify this if we collapse it into a single "serializer".


tests/__snapshots__/test_snapshots.yaml
```diff
@@ -36,12 +36,12 @@ test_set:
     this: null
 test_set.1:
   data: !!set
-    this: null
-    is: null
     a: null
     ? !!python/object/apply:builtins.frozenset
     - - nested, set
     : null
+    is: null
```
noahnu: yaml should be sorting the keys in alphabetical order, so having "tihs" and "is" after "a" is correct. Perhaps an issue with different pyaml versions, or a limitation of "set"?

src/syrupy/plugins/image/serializer.py
```diff
@@ -1,5 +1,52 @@
+from typing import Any, Set, Optional
```
noahnu: copy & pasted and then swapped "io" for "serializer". If I split this into 2 commits, maybe GitHub would pick up the diff better?

README.md
```diff
@@ -58,16 +58,15 @@ import pytest
 @pytest.fixture
 def snapshot_custom(snapshot):
     return snapshot.with_class(
-        io_class=CustomIOClass,
-        serializer_class=CustomSerializerClass,
+        serializer_class=CustomIOClass,
```
iamogbz: Shouldn't this be `CustomSnapshotSerializer`?
```suggestion
        serializer_class=CustomSnapshotSerializer,
```

README.md
```diff
@@ -58,16 +58,15 @@ import pytest
...
     )
 
 def test_image(snapshot_custom):
     actual = "..."
     assert actual == snapshot_custom
 ```
 
-Both `CustomIOClass` and `CustomSerializerClass` should extend `syrupy.io.SnapshotIO` and `syrupy.io.SnapshotSerializer` respectively.
+`CustomIOClass` should extend `syrupy.serializer.SnapshotSerializer`.
```
iamogbz:
```suggestion
`CustomSnapshotSerializer` should extend `syrupy.serializer.SnapshotSerializer`.
```

src/syrupy/assertion.py
```diff
@@ -15,14 +14,12 @@ def __init__(
...
         update_snapshots: bool,
-        io_class: Type[SnapshotIO],
         serializer_class: Type[SnapshotSerializer],
         test_location: TestLocation,
         session,
     ):
         self._update_snapshots = update_snapshots
-        self._io_class = io_class
-        self._serializer_class = serializer_class
+        self._io_class = serializer_class
```
iamogbz: And all the other renames
```suggestion
        self._serializer_class = serializer_class
```

src/syrupy/plugins/image/serializer.py
iamogbz: That's annoying. I think it's just the fact that this file existed before, so it does not register it as a move/rename

tests/__snapshots__/test_snapshots.yaml
iamogbz: Might be a set limitation not having a specific order? Hmm, also if you merge master, I think I fixed the issue where it rewrites an existing passing snapshot. You can revert this snapshot change.

iamogbz: Not sure why this is happening. Also thought I fixed the issue where it tries to rewrite existing passing snapshots

noahnu: I actually deleted this file and reran update snapshots

src/syrupy/assertion.py
```diff
@@ -91,12 +78,11 @@ def __eq__(self, other) -> bool:
     def _assert(self, data) -> bool:
         self._session.register_assertion(self)
         try:
-            serialized_data = self.serializer.encode(data)
```
iamogbz: I think we mistakenly reverted the checking without deserializing in this PR 🤦‍♂

PR #31, opened 2019-12-08
style: add strict typing

noahnu:
Uses forward references where possible. If we drop support for python 3.6, we can use implicit forward references via a future import. I think it's worth keeping 3.6 for now.

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -1,4 +1,5 @@
 import pytest
+from typing import Any, Optional, List
 
 from .assertion import SnapshotAssertion
 from .serializer import SnapshotSerializer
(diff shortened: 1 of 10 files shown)


src/syrupy/__init__.py
```diff
@@ -1,12 +1,13 @@
...
+from typing import Any, Optional, List
 
 from .assertion import SnapshotAssertion
 from .serializer import SnapshotSerializer
 from .location import TestLocation
 from .session import SnapshotSession
 
 
-def pytest_addoption(parser):
+def pytest_addoption(parser: Any) -> None:
```
noahnu: These pytest hooks can likely be updated to use proper types (can do that in a follow up).

PR #34, opened 2019-12-09
chore: sort imports

iamogbz:
## Description

Sort imports as part of formatting and reducing diff size

### Comments

1. Move `mypy.ini` into pyproject.toml once python/mypy#5205 is resolved

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,2 +1,3 @@
+# TODO: move to pyproject.toml once python/mypy#5205 is resolved
 [mypy]
 mypy_path=stubs
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,3 +1,2 @@
-# TODO: move to pyproject.toml once python/mypy#5205 is resolved
 [mypy]
 mypy_path=stubs
--- a/tasks.py
+++ b/tasks.py
@@ -24,7 +24,7 @@ def lint(ctx, fix=False):
     }
 
     for section, command in lint_commands.items():
-        print(f"\033[1m{section.capitalize()}\033[0m")
+        print(f"\033[1m{section}\033[0m")
         ctx.run(command or skip_command, pty=True)
         print()
 
--- a/tasks.py
+++ b/tasks.py
@@ -16,16 +16,15 @@ def requirements(ctx):
 
 @task
 def lint(ctx, fix=False):
-    skip_command = "echo 'Skipping'"
     lint_commands = {
+        "mypy": "python -m mypy --strict src",
         "isort": f"python -m isort {'' if fix else '--check-only --diff'} -y",
         "black": f"python -m black {'' if fix else '--check'} .",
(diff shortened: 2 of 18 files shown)


tasks.py
```diff
@@ -16,16 +16,17 @@ def requirements(ctx):
...
-        ctx.run("python -m mypy --strict src", pty=True)
+    skip_command = "echo 'Skipping'"
+    lint_commands = {
+        "isort": f"python -m isort {'' if fix else '--check-only --diff'} -y",
+        "black": f"python -m black {'' if fix else '--check'} .",
+        "mypy": None if fix else f"python -m mypy --strict src",
+    }
+
+    for section, command in lint_commands.items():
+        print(f"\033[1m{section.capitalize()}\033[0m")
```
noahnu:
```suggestion
        print(f"\033[1m{section}\033[0m")
```

capitalization doesn't make sense here

tasks.py
```diff
@@ -16,16 +16,17 @@ def requirements(ctx):
...
-    if fix:
-        print("\nSkipping type check as there is no fixer")
-    else:
-        print("\nRunning type check")
-        ctx.run("python -m mypy --strict src", pty=True)
+    skip_command = "echo 'Skipping'"
+    lint_commands = {
+        "isort": f"python -m isort {'' if fix else '--check-only --diff'} -y",
+        "black": f"python -m black {'' if fix else '--check'} .",
+        "mypy": None if fix else f"python -m mypy --strict src",
```
noahnu: We should run mypy always, otherwise you end up running the linter task twice, once to fix and once to check for mypy issues.

tasks.py
```diff
@@ -16,16 +16,17 @@ def requirements(ctx):
...
-
-    if fix:
-        print("\nSkipping type check as there is no fixer")
-    else:
-        print("\nRunning type check")
-        ctx.run("python -m mypy --strict src", pty=True)
+    skip_command = "echo 'Skipping'"
+    lint_commands = {
+        "isort": f"python -m isort {'' if fix else '--check-only --diff'} -y",
+        "black": f"python -m black {'' if fix else '--check'} .",
```
noahnu: interestingly black doesn't check for unused imports and the like. Do we need to add pylint for that?

mypy.ini
```diff
@@ -1,2 +1,3 @@
+# TODO: move to pyproject.toml once python/mypy#5205 is resolved
```
noahnu: we can remove the TODO and track this in a github issue

tasks.py
iamogbz: I wanted to, but relegated that to a different issue/pr

mypy.ini
iamogbz:
```suggestion
```

tasks.py
iamogbz: Sure, I was preserving the previous behaviour in here

Issue #23, opened 2019-12-06
More tests!


iamogbz:
**Is your feature request related to a problem? Please describe.**

No integration tests mean wide ranging can introduce bugs and regressions not caught.

**Describe the solution you'd like**

💯 coverage!

**Describe alternatives you've considered**

No bragging rights for 💯 coverage

**Additional context**

N/A

PR #35, opened 2019-12-11
ci: set up coverage

iamogbz:
Sets up code coverage reporting in prep for #23

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -22,12 +22,13 @@ jobs:
     - name: Install python dev dependencies.
       run: |
         python -m pip install -U pip --no-cache-dir
-        python -m pip install pip-tools wheel
         python -m pip install -e . -r requirements.txt -r dev-requirements.txt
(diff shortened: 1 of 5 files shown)


.github/workflows/pythonapp.yml
```diff
@@ -22,12 +22,13 @@ jobs:
...
       run: |
         python -m pip install -U pip --no-cache-dir
-        python -m pip install pip-tools wheel
         python -m pip install -e . -r requirements.txt -r dev-requirements.txt
     - name: Lint
       run: invoke lint
     - name: Test
-      run: invoke test
+      env:
+        CODECOV_TOKEN: 282b0064-e377-478b-834c-b8de1a3ee4c4
```
noahnu: Should the token be committed or in a secret?

iamogbz: It can be replaced with a secret, I would. However it's not really sensitive as the worst that can happen is someone uploads a false coverage report.

Issue #22, opened 2019-12-06
Pretty diff of snapshot comparisons


iamogbz:
**Is your feature request related to a problem? Please describe.**

Diffing complex objects on assertion failure does not show only the change but the entire snapshot

**Describe the solution you'd like**

Should show the exact line and characters (tokens) that differ between snapshots

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

noahnu:
This should be done on the same raw string that's written to the file system, however we should also support custom types. E.g. how to pretty diff images via the image serializer plugin?

noahnu:
Only thing we're missing here is showing exact characters within a line. Should we only do this for single line diffs? I imagine it might be too verbose in the general case.

PR #46, opened 2019-12-22
Compare serialized data in assertion compr & improve diff report

noahnu:
Partially addresses #22 by spliting diff reporting line by line, similar to jest snapshot diffs.

Addresses #21 by reading raw text from the snapshot file for line by line comparison.


.coveragerc
```diff
@@ -0,0 +1,4 @@
+[report]
```
noahnu: not sure how to put this in the pyproject.toml

tests/test_snapshots.py
```diff
@@ -39,7 +39,6 @@ def test_dict(snapshot, actual):
 
 def test_set(snapshot):
     assert snapshot == {"this", "is", "a", "set"}
-    assert snapshot == {"this", "is", "a", frozenset({"nested, set"})}
```
noahnu: A frozenset does not have a guaranteed order when it's serialized in pyyaml. A "set" itself can also not contain a set. I don't think this test adds much value in that case.

tests/test_image_plugin.py
```diff
@@ -19,3 +19,8 @@ def test_image(snapshot_png):
         b"RK5CYII="
     )
     assert actual == snapshot_png
+
+
+def test_raises_error_for_unserializable_data(snapshot_png):
+    with pytest.raises(ValueError):
+        assert "not a byte string" == snapshot_png
```
noahnu: not sure if it's best to throw an error here, or consider this a failed snapshot.

src/syrupy/serializers/yaml.py
```diff
@@ -68,3 +81,31 @@ def _read_file(self, filepath: str) -> Any:
...
+                    if not indent:
+                        if cur_name is not None:
+                            raw_snapshots[cur_name] = cur_data
+
+                        cur_name = line[:-2]  # newline & colon
+                        cur_data = ""
+                    else:
+                        cur_data = f"{cur_data}{line[2:]}"
+                if cur_name is not None:
+                    raw_snapshots[cur_name] = cur_data
```
iamogbz: I think this can be simplified by not having the `cur_data` staging
```py
        raw_snapshots = {}
        cur_name = None
        try:
            with open(filepath, "r") as f:
                for line in f:
                    indent = len(line) - len(line.lstrip(" "))
                    if not indent:
                        cur_name = line[:-2]  # newline & colon
                        raw_snapshots[cur_name] = ""
                    else:
                        raw_snapshots[cur_name] += line[2:]
```

tests/test_snapshots.py
iamogbz: Discussed offline, this is a limitation with yaml as a snapshot library should be able to handle any arbitrary data type

tests/test_image_plugin.py
iamogbz: We could wrap the serialization with an error handler and compare the raw values if it fails

tasks.py
```diff
@@ -59,14 +59,17 @@ def test(ctx, coverage=False, dev=False, update_snapshots=False, verbose=False):
     """
     env = {"PYTHONPATH": "./src"} if dev else {}
     flags = {
-        "-s": verbose,
+        "-s -vv": verbose,
```
iamogbz: 👍

src/syrupy/assertion.py
```diff
@@ -85,11 +91,31 @@ def assert_match(self, data: "SerializableData") -> None:
...
         if snapshot_data is None:
             return ["Snapshot does not exist!"]
 
         if not assertion_result.success:
-            return [f"- {data}", f"+ {snapshot_data}"]
+            received = serialized_data.splitlines()
+            stored = snapshot_data.splitlines()
+
+            marker_stored = bold(green("-"))
+            marker_received = bold(red("+"))
```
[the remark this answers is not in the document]
iamogbz: Can add `success_style` for `bold(green(...))` I think we use it elsewhere

.coveragerc
iamogbz: Issue resolved here: https://github.com/nedbat/coveragepy/issues/664
Documentation here: https://coverage.readthedocs.io/en/coverage-5.0/config.html

pyproject.toml
```diff
@@ -41,3 +41,10 @@ disable = '''
 missing-function-docstring,
 missing-module-docstring,
 '''
+
+[tool.coverage.report]
+exclude_lines = [
```
iamogbz: Can also remove import lines, method name and decorator lines?

PR #60, opened 2019-12-25
fix: only run updates on selected test nodes

### LINKED ISSUE #19: Deletes file when running only subset of tests in test file with update snapshots flag
**Describe the bug**

The snapshot file holding the serialized assertions gets deleted when running in partial mode and update snapshots enabled

**To Reproduce**
1. Clone repo and `. script/bootstrap`
1. `inv test -du -t=test_multiple_snapshots`
1. ~This deletes the entire file~ fixed in #20 
1. This deletes all files and snapshots not touched by the subset of tests

**Expected behavior**

1. Only deletes unused snapshots and preserves newly written snapshots
1. Only modifies files to be touched by a subset of the test

**Additional context**

iamogbz:
## Description

Resolves #19 

### QA

See issue str

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -13,7 +13,10 @@ from .session import SnapshotSession
 
 
 def pytest_addoption(parser: Any) -> None:
-    """Exposes snapshot plugin configuration to pytest."""
+    """
+    Exposes snapshot plugin configuration to pytest.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_addoption
+    """
     group = parser.getgroup("syrupy")
     group.addoption(
         "--snapshot-update",
@@ -25,6 +28,10 @@ def pytest_addoption(parser: Any) -> None:
 
 
 def pytest_assertrepr_compare(op: str, left: Any, right: Any) -> Optional[List[str]]:
+    """
+    Return explanation for comparisons in failing assert expressions.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_assertrepr_compare
+    """
     if isinstance(left, SnapshotAssertion):
         assert_msg = f"snapshot {op} {right}"
         return [assert_msg] + left.get_assert_diff(right)
@@ -35,6 +42,10 @@ def pytest_assertrepr_compare(op: str, left: Any, right: Any) -> Optional[List[s
 
 
 def pytest_sessionstart(session: Any) -> None:
+    """
+    Initialize snapshot session before tests are collected and ran.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_sessionstart
+    """
     config = session.config
     session._syrupy = SnapshotSession(
         update_snapshots=config.option.update_snapshots, base_dir=config.rootdir
@@ -43,14 +54,26 @@ def pytest_sessionstart(session: Any) -> None:
 
 
 def pytest_collection_modifyitems(session: Any, config: Any, items: List[Any]) -> None:
+    """
+    After tests are collected and before any modification is performed.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_modifyitems
+    """
     session._syrupy._all_items.update(items)
 
 
 def pytest_collection_finish(session: Any) -> None:
+    """
+    After collection has been performed and modified.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_finish
+    """
     session._syrupy._ran_items.update(session.items)
 
 
 def pytest_sessionfinish(session: Any) -> None:
+    """
+    Add syrupy report to pytest after whole test run finished, before exiting.
+    https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_sessionfinish
+    """
     reporter = session.config.pluginmanager.get_plugin("terminalreporter")
     session._syrupy.finish()
     for line in session._syrupy.report:
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -42,6 +42,14 @@ def pytest_sessionstart(session: Any) -> None:
     session._syrupy.start()
 
 
+def pytest_collection_modifyitems(session: Any, config: Any, items: List[Any]) -> None:
+    session._syrupy._all_items.update(items)
+
+
+def pytest_collection_finish(session: Any) -> None:
+    session._syrupy._ran_items.update(session.items)
+
+
 def pytest_sessionfinish(session: Any) -> None:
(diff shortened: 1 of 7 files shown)


src/syrupy/serializers/base.py
```diff
@@ -93,10 +116,21 @@ def post_read(self, index: int = 0) -> None:
     def pre_write(self, data: "SerializableData", index: int = 0) -> None:
         self._ensure_snapshot_dir(index)
 
+    @final
     def write(self, data: "SerializableData", index: int = 0) -> None:
+        """
+        Override _write_snapshot_or_remove_file in subclass to change behaviour
```
iamogbz: This method should not be overridden but rather use `get_filepath`, `get_snapshot_name` and the internal `_write_snapshot_or_remove_file` which can be overridden to customise write behaviour

src/syrupy/serializers/base.py
```diff
@@ -70,19 +88,24 @@ def create_or_update_snapshot(self, data: "SerializableData", index: int) -> Non
...
-        self.write_snapshot_or_remove_file(snapshot_file, snapshot_name, None)
+        self._write_snapshot_or_remove_file(snapshot_file, snapshot_name, None)
 
     def pre_read(self, index: int = 0) -> None:
         pass
 
+    @final
     def read(self, index: int = 0) -> "SerializableData":
+        """
+        Override _read_snapshot_from_file in subclass to change behaviour
```
iamogbz: This method should not be overridden but rather use `get_filepath`, `get_snapshot_name` and the internal `_read_snapshot_from_file` which can be overridden to customise read behaviour

src/syrupy/serializers/base.py
```diff
@@ -93,10 +116,21 @@ def post_read(self, index: int = 0) -> None:
...
+        Override _write_snapshot_or_remove_file in subclass to change behaviour
+        """
         snapshot_file = self.get_filepath(index)
         snapshot_name = self.get_snapshot_name(index)
-        self.write_snapshot_or_remove_file(snapshot_file, snapshot_name, data)
+        if self.test_location.testname not in snapshot_name:
+            print(
+                warning_style(
+                    f"Warning: snapshot name '{snapshot_name}' does not"
+                    f" contain testname '{self.test_location.testname}'"
```
iamogbz: Added because this relies on the test names for deciding which tests were run by pytest collection

src/syrupy/serializers/base.py
noahnu: "Private" methods aren't part of the public API and so shouldn't be overridden by derived classes. Last time I refactored the abstract class, I explicitly renamed functions to remove the leading underscore if they were overwritten.

src/syrupy/__init__.py
```diff
@@ -42,6 +42,14 @@ def pytest_sessionstart(session: Any) -> None:
     session._syrupy.start()
 
 
+def pytest_collection_modifyitems(session: Any, config: Any, items: List[Any]) -> None:
```
noahnu: let's link the pytest docs: https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_modifyitems

src/syrupy/__init__.py
```diff
@@ -42,6 +42,14 @@ def pytest_sessionstart(session: Any) -> None:
...
+    session._syrupy._all_items.update(items)
+
+
+def pytest_collection_finish(session: Any) -> None:
```
noahnu: https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_finish

src/syrupy/__init__.py
```diff
@@ -42,6 +42,14 @@ def pytest_sessionstart(session: Any) -> None:
...
+    session._syrupy._ran_items.update(session.items)
+
+
 def pytest_sessionfinish(session: Any) -> None:
```
noahnu: https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_sessionfinish

src/syrupy/__init__.py
noahnu: If we use https://github.com/Instagram/MonkeyType, can we avoid the "Any" types?

src/syrupy/serializers/base.py
```diff
@@ -40,6 +43,19 @@ def dirname(self) -> str:
             return os.path.join(test_dirname, SNAPSHOT_DIRNAME, subdir_name)
         return os.path.join(test_dirname, SNAPSHOT_DIRNAME)
 
+    @property
+    def snapshot_subdirectory_name(self) -> Optional[str]:
```
noahnu: this is just grouping methods/fields that should be overridden together?

src/syrupy/serializers/base.py
```diff
@@ -133,7 +162,7 @@ def _ensure_snapshot_dir(self, index: int) -> None:
             pass
 
     @abstractmethod
-    def read_snapshot_from_file(
+    def _read_snapshot_from_file(
```
noahnu: mentioned above, I prefer removing the underscore if we want to allow subclasses to override

src/syrupy/session.py
```diff
@@ -42,12 +44,16 @@ def __init__(self, *, update_snapshots: bool, base_dir: str):
         self.update_snapshots = update_snapshots
         self.base_dir = base_dir
         self.report: List[str] = []
+        self._all_items: Set[Any] = set()
```
noahnu: Any?

src/syrupy/session.py
```diff
@@ -205,3 +219,7 @@ def _diff_snapshot_files(
 
     def _count_snapshots(self, snapshot_files: "SnapshotFiles") -> int:
         return sum(len(snaps) for snaps in snapshot_files.values())
+
+    @staticmethod
+    def get_node_testname(node: Any) -> str:
```
noahnu: we can probably find the name of the class and use a forward ref.

src/syrupy/serializers/base.py
iamogbz: Yeah essentially

src/syrupy/serializers/base.py
iamogbz: I had meant to call out that these methods should not be accessed externally on the object instance but can still be overridden. By convention double underscore methods should not be overridden though.

iamogbz: https://stackoverflow.com/a/14898239/3072906

src/syrupy/__init__.py
iamogbz: done

src/syrupy/__init__.py
iamogbz: done

src/syrupy/serializers/base.py
iamogbz: We can do that with double underscore methods, and still use underscore methods for internal class methods that can be overridden

PR #62, opened 2019-12-27
fix: yaml multiline string should be read successfully

noahnu:
The bug this fixes:

Adding

```
def test_multiline(snapshot):
    assert "multi\nline" == snapshot
```

to its own test file and running `inv test -u` would create a corrupted file.

### DIFF
--- a/src/syrupy/serializers/base.py
+++ b/src/syrupy/serializers/base.py
@@ -88,12 +88,13 @@ class AbstractSnapshotSerializer(ABC):
         self.write(data, index=index)
         self.post_write(data, index=index)
 
-    @final
+    @abstractmethod
     def delete_snapshot_from_file(self, snapshot_file: str, snapshot_name: str) -> None:
         """
-        Utility method for removing a snapshot from a snapshot file.
+        Remove a snapshot from a snapshot file.
+        If the snapshot file will be empty remove the entire file.
         """
-        self._write_snapshot_or_remove_file(snapshot_file, snapshot_name, None)
+        raise NotImplementedError
 
     def pre_read(self, index: int = 0) -> None:
         pass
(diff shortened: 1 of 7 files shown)


noahnu:
This is also an argument against custom parsers.

tests/test_snapshots.py
```diff
@@ -12,6 +12,15 @@ def test_simple_string(snapshot):
     assert "Loreeeeeem ipsum." == snapshot
 
 
+def test_multiline(snapshot):
+    assert "multi\nline" == snapshot
+
+
+def test_singleline(snapshot):
+    assert "one-line" == snapshot
```
iamogbz: Not sure what this assertion adds, we already test single lines above. Is it the `-`?

iamogbz: Should also add a test case for python multi lines `"""..."""`

tests/test_snapshots.py
```diff
@@ -12,6 +12,15 @@ def test_simple_string(snapshot):
...
+    assert "multi\nline 2\nline 3" == snapshot
```
iamogbz: Should also add a test case for python multi lines `"""..."""`

src/syrupy/serializers/base.py
```diff
@@ -88,12 +88,13 @@ def create_or_update_snapshot(self, data: "SerializableData", index: int) -> Non
         self.write(data, index=index)
         self.post_write(data, index=index)
 
-    @final
+    @abstractmethod
     def delete_snapshot_from_file(self, snapshot_file: str, snapshot_name: str) -> None:
```
iamogbz: This is now an explicit removal method

src/syrupy/serializers/raw_single.py
```diff
@@ -44,13 +44,13 @@ def _read_file(self, filepath: str) -> Any:
...
 
-    def _write_snapshot_or_remove_file(
+    def _write_snapshot_to_file(
         self, snapshot_file: str, snapshot_name: str, data: "SerializableData"
     ) -> None:
-        if data:
-            self.__write_file(snapshot_file, data)
-        else:
-            os.remove(snapshot_file)
+        self.__write_file(snapshot_file, data)
```
iamogbz: Alway write snapshot data even if `None`

noahnu:
Changes LGTM

PR #64, opened 2019-12-27
fix: update on matching parameterized test snapshots

### LINKED ISSUE #63: Partial test run snapshot update does not support pytest parametrize
**Describe the bug**

When running a subset of all tests with the pytest keyword flag `-k`,  snapshots in tests matching the keyword should be updated and deleted appropriately. This does not work for `pytest.mark.parametrize` tests

**To Reproduce**
1. Checkout master
1. Remove a parametrize test case e.g.
https://github.com/tophat/syrupy/blob/7cde8a7b0773d3ecbf6b8e1a8d677ba8c982dfa2/tests/test_snapshots.py#L37
1. Run `inv test -d -u -t=<test_name>`
1. Test case does not get removed

**Expected behavior**

Test case removed from the parametrized definition gets deleted

**Screenshots**
N/A

 - OS: macos catalina
 - Version: a15276c

**Additional context**
N/A

iamogbz:
**Description**

Fixes #63 
Fixes removing hanging snapshot file when all tests have been removed

**QA**

See issue str


src/syrupy/session.py
```diff
@@ -176,19 +176,22 @@ def _collate_snapshots(self) -> None:
...
-                else {
+            if ran_all:
+                unused = unused_snapshots
+            else:
+                unused = {
                     snapshot_name
                     for snapshot_name in unused_snapshots
-                    if any({testname in snapshot_name for testname in ran_testnames})
+                    if any(
+                        {
```
noahnu: unnecessary set. Should replace with a generator:
```
any(self.snapshot_... for node in self._ran_items)
```

src/syrupy/session.py
```diff
@@ -216,5 +219,7 @@ def _count_snapshots(self, snapshot_files: "SnapshotFiles") -> int:
         return sum(len(snaps) for snaps in snapshot_files.values())
 
     @staticmethod
-    def get_node_testname(node: Any) -> str:
-        return str(getattr(node, "name", node.obj.__name__ if node.obj else ""))
+    def snapshot_name_matches_test_node(name: str, node: Any) -> bool:
+        methodname = node.obj.__name__ if node.obj else None
+        nodename = getattr(node, "name", None)
+        return methodname in name or name in nodename
```
noahnu: this will produce false positives for names like: `test_case_1` and `test_case_10`.

iamogbz: Not sure to disambiguate this, since `test_case_1` and `test_case_10` can be generated by `test_case_1` with the same serializer

src/syrupy/__init__.py
```diff
@@ -82,17 +82,9 @@ def pytest_sessionfinish(session: Any) -> None:
...
-        classname=request.cls.__name__ if request.cls else None,
-        methodname=request.function.__name__ if request.function else None,
-        nodename=getattr(request.node, "name", ""),
-        testname=SnapshotSession.get_node_testname(request.node),
-    )
     return SnapshotAssertion(
         update_snapshots=request.config.option.update_snapshots,
         serializer_class=DEFAULT_SERIALIZER,
-        test_location=test_location,
+        test_location=TestLocation(request=request),
```
iamogbz: Can easily be refactored to use the `request.node` later on

tests/test_integration.py
```diff
@@ -222,6 +264,25 @@ def test_removed_empty_snapshot_file_only(stubs):
...
     assert os.path.relpath(filepath) not in result_stdout
     assert "1 snapshot deleted" in result_stdout
+    assert "empty snapshot file" in result_stdout
+    assert os.path.relpath(empty_filepath) in result_stdout
     assert result.ret == 0
     assert os.path.isfile(filepath)
     assert not os.path.isfile(empty_filepath)
+
+
+def test_removed_hanging_snapshot_file(stubs):
```
iamogbz: Test that snapshot dir gets cleaned out when running all tests

src/syrupy/session.py
iamogbz: Leaving the logic as is, the issue is there's no way to guarantee the serializer still exists, if it does then good we can use it to remove the snapshot. If it does not then we have to still detect that the same function no longer asserts on those snapshots when ran. I think the assumption that method names would match snapshot names to some degree is safe.

src/syrupy/location.py
```diff
@@ -1,7 +1,29 @@
...
+    def __init__(self, node: Any):
+        self._node = node
+        self.filename = self._node.fspath
+        self.modulename = self._node.obj.__module__
+        self.methodname = self._node.obj.__name__
+        self.nodename = getattr(self._node, "name", None)
+        self.testname = self.nodename or self.methodname
+
+    @property
+    def classname(self) -> Optional[str]:
```
iamogbz: Not sure if there's a better way to extract the class name from the `request.node` when used in `session.py#196` the `node` does have a `request` object attached

src/syrupy/session.py
```diff
@@ -138,10 +140,11 @@ def remove_unused_snapshots(
...
         for snapshot_file, unused_snapshots in unused_snapshot_files.items():
-            for snapshot_name in unused_snapshots:
-                self._serializers[snapshot_file].delete_snapshot_from_file(
-                    snapshot_file, snapshot_name
-                )
+            serializer = self._serializers.get(snapshot_file)
+            if serializer:
+                serializer.delete_snapshots_from_file(snapshot_file, unused_snapshots)
+            elif snapshot_file not in used_snapshot_files:
+                os.remove(snapshot_file)
```
noahnu: I thought this has been moved into the serializer?

iamogbz: If the serializer has been removed, there needs to be a way to remove unused files. I forgot that before in your PR

PR #68, opened 2019-12-27
feat: remove call syntax, close #54

### LINKED ISSUE #54: Remove assertion call syntax

One of the project's motivations is to provide a uniform, idiomatic way to assert snapshots. We're keeping `snapshot.assert_match` for backwards compatibility with snapshottest for the moment, however no one uses the `snapshot(actual)` syntax. It's a remnant of testing out different approaches.

Remove the snapshot call syntax: https://github.com/tophat/syrupy/blob/7cc40476556cbb7ddb3f60bab81668341550a742/src/syrupy/assertion.py#L123

noahnu:


tests/test_snapshots.py
```diff
@@ -35,7 +35,6 @@ def test_string(snapshot, actual):
 def test_multiple_snapshots(snapshot):
     assert "First." == snapshot
     snapshot.assert_match("Second.")
-    snapshot("Third.")
```
iamogbz: This could be just changed to `snapshot == "Third."

PR #69, opened 2019-12-27
feat: use custom serializer instead of pyyaml

### LINKED ISSUE #38: Make pyyaml dependency optional

Syrupy supports multiple serializes including custom serializes. If a codebase using syrupy selects a non-yaml serializes, there is no reason to install the unused dependency.

List pyyaml as an optional dependency. Throw error if the seralizer is used if pyyaml is not present in system.

- Continue installing unused dependencies.

**Additional context**

Python dependency management is awful :smile:

noahnu:
close #38 , removes yaml

## Known Issues

- Serializing classes which do not provide a repr/str function will include memory addresses in the serialized representation.
- Cyclic data structures will not terminate at some max depth.


src/syrupy/serializers/amber.py
```diff
@@ -0,0 +1,216 @@
...
+    ---
+    # name: test_name_2
+
+     data
+    ```
+    """
+
+    @property
+    def file_extension(self) -> str:
+        return "snap"
```
iamogbz: AmBeR!!!
```suggestion
        return "ambr"
```

src/syrupy/serializers/amber.py
```diff
@@ -0,0 +1,216 @@
...
+            for key in sorted(snapshots.keys()):
+                snapshot = snapshots[key]
+                snapshot_data = snapshot.get("data")
+                if snapshot_data is not None:
+                    f.write(f"{self.__name_marker} {key}\n")
+                    for data_line in snapshot_data.split("\n"):
+                        f.write(f"{self.__indent}{data_line}\n")
+                    f.write(f"{self.__divider}\n")
+
+    def __read_file(self, filepath: str) -> Dict[str, Dict[str, Any]]:
```
iamogbz: Shouldn't this logic live in the `DataSerializer`

tests/test_amber_serializer.py
```diff
@@ -51,6 +51,7 @@ def test_dict(snapshot, actual):
 
 def test_set(snapshot):
     assert snapshot == {"this", "is", "a", "set"}
+    assert snapshot == {"contains", "frozen", frozenset({"1", "2"})}
```
iamogbz: Nice!

iamogbz: Also file can be renamed back to `tests/test_snapshots.py`

tests/__snapshots__/test_amber_serializer.ambr
```diff
@@ -0,0 +1,137 @@
...
+    'this',
+    'is',
+    <class 'tuple'> (
+      'a',
+      'tuple',
+    ),
+  )
+---
+# name: test_tuple.1
+  <class 'tests.test_amber_serializer.ExampleTuple'> (
```
iamogbz: To prevent brittleness, we could use just the class name

tests/test_amber_serializer.py
noahnu: actually renaming the "plugin" tests to `test_image_serializer` format

README.md
```diff
@@ -46,6 +46,15 @@ pytest --snapshot-update
 
 A snapshot file should be generated under a `__snapshots__` directory in the same directory as `test_file.py`. The `__snapshots__` directory and all its children should be committed along with your test code.
 
+### Serializers
+
+Syrupy comes with a few built-in serializers for you to choose from. You should also feel free to extend the AbstractSnapshotSerializer if your project has a need not captured by one our built-ins.
+
+- AmberSnapshotSerializer: This is the default serializer which generates `.ambr` files. Serialization of most data types are supported, however non-sortable types such as frozenset are experimental.
```
iamogbz:
```suggestion
- **AmberSnapshotSerializer**: This is the default serializer which generates `.ambr` files. Serialization of most data types are supported, however non-sortable types such as frozenset are experimental.
```

README.md
```diff
@@ -46,6 +46,15 @@ pytest --snapshot-update
...
+- RawSingleSnapshotSerializer: Unlike the AmberSnapshotSerializer, which groups all tests within a single test file into a singular snapshot file, the Raw Single serializer creates one `.raw` file per test case.
```
iamogbz:
```suggestion
- **`RawSingleSnapshotSerializer`**: Unlike the `AmberSnapshotSerializer`, which groups all tests within a single test file into a singular snapshot file, the Raw Single serializer creates one `.raw` file per test case.
```

README.md
```diff
@@ -46,6 +46,15 @@ pytest --snapshot-update
...
+- PNGSnapshotSerializer: An extension of the Raw Single serializer, this should be used to produce `.png` files.
```
iamogbz:
```suggestion
- **`PNGSnapshotSerializer`**: An extension of the Raw Single serializer, this should be used to produce `.png` files.
```

README.md
```diff
@@ -46,6 +46,15 @@ pytest --snapshot-update
...
+- SVGSnapshotSerializer: Another extension of Raw Single. This produces `.svg` files from an svg string.
```
iamogbz:
```suggestion
- **`SVGSnapshotSerializer`**: Another extension of Raw Single. This produces `.svg` files from an svg string.
```

PR #78, opened 2019-12-29
fix: exit serialization early on detection of a cycle

### LINKED ISSUE #72: Serializer should fail on max depth exceeded to break cycles
**Describe the bug**

If you serialize an object containing a cycle, it will recurse until call stack overflows / python can no longer handle it, rather than a graceful application error.

**To Reproduce**

```
cycle = { "a": {} }
cycle["a"]["b"] = cycle

def test_cycle(snapshot):
    assert snapshot == cycle
```

**Expected behaviour**

Error should be thrown if a cycle is detected.

**Additional context**

We already track "indent". We could add some logic to throw an error if indent exceeds some level as it's essentially equivalent to depth.

iamogbz:
## Description

Tracks visited items in serialization and exits early when cycle is encountered or max depth exceeded

## Related Issues

- Closes #72 

## Checklist

## Additional Comments

- Opted to not fail on cycle, since this is not necessarily an invalid data structure


src/syrupy/serializers/amber.py
```diff
@@ -129,31 +147,52 @@ def serialize_dict(cls, data: "SerializableData", indent: int = 0) -> str:
...
+    def serialize_unknown(
+        cls, data: Any, *, indent: int = 0, visited: List[Any] = []
+    ) -> str:
+        return cls.with_indent(repr(data), indent)
+
+    @classmethod
+    def serialize(
+        cls, data: "SerializableData", *, indent: int = 0, visited: List[Any] = []
+    ) -> str:
+        if indent > cls._max_depth or data in visited:
```
noahnu: we should rename indent to `depth`

src/syrupy/serializers/amber.py
```diff
@@ -129,31 +147,52 @@ def serialize_dict(cls, data: "SerializableData", indent: int = 0) -> str:
...
     @classmethod
-    def serialize(cls, data: "SerializableData", indent: int = 0) -> str:
+    def serialize_unknown(
+        cls, data: Any, *, indent: int = 0, visited: List[Any] = []
+    ) -> str:
+        return cls.with_indent(repr(data), indent)
+
+    @classmethod
+    def serialize(
+        cls, data: "SerializableData", *, indent: int = 0, visited: List[Any] = []
```
noahnu: should use a Set instead of List

src/syrupy/serializers/amber.py
```diff
@@ -89,7 +92,9 @@ def object_type(cls, data: "SerializableData") -> str:
         return f"<class '{data.__class__.__name__}'>"
 
     @classmethod
-    def serialize_string(cls, data: "SerializableData", indent: int = 0) -> str:
+    def serialize_string(
+        cls, data: "SerializableData", *, indent: int = 0, visited: List[Any] = []
```
noahnu: we can remove the unused `visited` kwarg if we do `**_`, not sure how mypy handles that though?

src/syrupy/serializers/amber.py
```diff
@@ -17,6 +18,8 @@
 
 
 class DataSerializer:
+    _max_depth: int = 99
+    _max_depth_value: str = "..."
     indent: str = "  "
```
noahnu: the other properties should also be prefixed with underscore for consistency

src/syrupy/serializers/amber.py
iamogbz: I think that requires each object to be hashable, which might not be feasible

src/syrupy/serializers/amber.py
iamogbz: I think it complained last I tried, will attempt again but treating as non blocking

src/syrupy/serializers/amber.py
noahnu: You could store the id in the set

iamogbz: The id from hash?

noahnu: Id function. Basically store the memory addresses in a set since they're integers and thus always hashable. We're keeping everything in memory anyway

iamogbz: Good point

iamogbz: Updated

src/syrupy/serializers/amber.py
iamogbz: Too much refactoring needed to satisfy mypy and ignore the argument, leaving as is

PR #80, opened 2019-12-29
chore: add flake8 plugins, comprehension, bugbear, builtins

noahnu:
## Description

Adds:
- https://github.com/adamchainz/flake8-comprehensions
- https://github.com/PyCQA/flake8-bugbear
- https://github.com/gforcada/flake8-builtins

## Checklist

- [x] Licenses of new dependencies are compatible with Apache 2.0.

## Additional Comments

The comprehension rule provides a minuscule performance benefit, however it ultimately brings us closer to consistently using "one pattern" when two are available (consistent code).

### DIFF
--- a/src/syrupy/serializers/amber.py
+++ b/src/syrupy/serializers/amber.py
@@ -6,6 +6,7 @@ from typing import (
     Any,
     Dict,
     Iterable,
+    Optional,
     Set,
 )
 
@@ -95,7 +96,11 @@ class DataSerializer:
 
     @classmethod
     def serialize_string(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
     ) -> str:
         if "\n" in data:
             return (
@@ -107,13 +112,21 @@ class DataSerializer:
 
     @classmethod
     def serialize_number(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
     ) -> str:
         return cls.with_indent(repr(data), depth)
 
     @classmethod
     def serialize_set(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
     ) -> str:
         return (
             cls.with_indent(f"{cls.object_type(data)} {{\n", depth)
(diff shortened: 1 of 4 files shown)


src/syrupy/serializers/amber.py
```diff
@@ -163,19 +184,28 @@ def serialize_iterable(
...
     @classmethod
     def serialize(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
     ) -> str:
+        visited = visited if visited is not None else set()
```
iamogbz: Wouldn't this also work?
```suggestion
        visited = visited or set()
```

src/syrupy/serializers/amber.py
```diff
@@ -208,7 +238,7 @@ def file_extension(self) -> str:
         return "ambr"
 
     def discover_snapshots(self, filepath: str) -> Set[str]:
-        return set(name for name in DataSerializer.read_file(filepath).keys())
+        return set(DataSerializer.read_file(filepath).keys())
```
iamogbz: 👍

src/syrupy/serializers/amber.py
```diff
@@ -126,9 +139,13 @@ def serialize_set(
...
     def serialize_dict(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
     ) -> str:
-        kwargs = dict(depth=depth + 1, visited=visited)
+        kwargs = {"depth": depth + 1, "visited": visited}
```
iamogbz: Why the change?

src/syrupy/serializers/amber.py
```diff
@@ -146,7 +163,11 @@ def serialize_dict(
 
     @classmethod
     def serialize_iterable(
-        cls, data: "SerializableData", *, depth: int = 0, visited: Set[Any] = set()
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
```
iamogbz: Why the refactor?

src/syrupy/serializers/amber.py
noahnu: no need to create a new set if the set is empty

src/syrupy/serializers/amber.py
noahnu: `{}` is preferred over `dict()`. It's slightly faster. Also more lint rules ensures we're consistent (two developers write same code)

src/syrupy/serializers/amber.py
noahnu: Previously we set the default arg to set(). This does not work like you'd expect. It doesn't create a "set" every function invocation, but rather re-uses a single set

PR #84, opened 2019-12-31
fix: get text everything

iamogbz:
## Description

Use `gettext` anywhere we show information to the syrupy plugin user

## Related Issues

N/A

## Checklist

## Additional Comments

- Changes made by running `inv requirements`

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -1,4 +1,3 @@
-from gettext import gettext
 from typing import (
     Any,
     List,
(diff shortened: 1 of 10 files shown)


src/syrupy/__init__.py
```diff
@@ -39,11 +40,12 @@ def pytest_assertrepr_compare(op: str, left: Any, right: Any) -> Optional[List[s
     Return explanation for comparisons in failing assert expressions.
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_assertrepr_compare
     """
+    snapshot_text = gettext("snapshot")
     if isinstance(left, SnapshotAssertion):
-        assert_msg = f"snapshot {op} {right}"
+        assert_msg = f"{snapshot_text} {op} {right}"
```
noahnu: This should be the same as the fixture name. Doesn't make sense to translate.

iamogbz: Updated to allow custom names

PR #87, opened 2020-01-03
build: remove requirement file

iamogbz:
## Description

Removes duplication and fragmentation of package dependencies in `requirements.txt` and `setup.py` file

## Related Issues

N/A

## Checklist

## Additional Comments
 
- One dependency list to rule them all


setup.py
```diff
@@ -8,6 +8,27 @@
...
+    "flake8-i18n",
+    "invoke",
+    "isort",
+    "mypy",
+    "pip-tools",
+    "py-githooks",
+    "pytest",
+    "semver",
+    "twine",
+    "wheel",
```
noahnu: What's the reasoning behind removing the version requirements?

iamogbz: The main purpose is to remove the duplication and fragmentation. The version pinning for dev dependencies (running tests) are not critical compared to the distribution requirements (e.g. typing_extensions)

iamogbz: We could pin them in here as well, just didn't think it was necessary

setup.py
```diff
@@ -8,6 +8,27 @@
 
 
 python_requires = "~=3.6"
+setup_requires = ["setuptools_scm"]
+install_requires = ["typing_extensions>=3.6"]
+dev_requires = [
+    "black==19.10b0",
```
noahnu: The issue with this pinning is that we don't have an easy way to update it. Our pip-tools command had an update flag.

iamogbz: Was doing some reading https://snarky.ca/clarifying-pep-518/

tasks.py
```diff
@@ -20,11 +20,15 @@ def requirements(ctx, upgrade=False):
...
     """
-    source = "requirements.in"
+    dest = "requirements.txt"
     args = ["--no-emit-find-links", "--no-index", "--allow-unsafe", "--rebuild"]
     if upgrade:
         args.append("--upgrade")
-    ctx.run(f"python -m piptools compile {source} {' '.join(args)}", pty=True)
+    ctx.run(
+        "echo '-e .[dev]' | python -m piptools compile "
+        f"{' '.join(args)} - -qo- | sed '/^-e / d' > {dest}",
```
iamogbz: Workaround from https://github.com/jazzband/pip-tools/issues/908

PR #94, opened 2020-01-10
chore: isolate linting, tests, and build

noahnu:
## Description

This PR splits up the linting, test, and build process into 3 distinct jobs. The main benefit here is the installation of the minimum number of dependencies when running tests, such that we don't have tests pass due to transitive dependencies, even if not declared a direct dependency.


setup.py
```diff
@@ -64,7 +66,7 @@ def readme() -> str:
         packages=find_packages("./src"),
         zip_safe=False,
         entry_points={"pytest11": ["syrupy = syrupy"]},
-        extras_require={"dev": dev_requires},
+        extras_require={"dev": dev_requires + test_requires, "test": test_requires},
```
iamogbz: There's a `tests_require` that provides this not sure how it'll work with `pip install`

noahnu: From https://setuptools.readthedocs.io/en/latest/setuptools.html:

> When you run the test command, setuptools will attempt to obtain these (using pip if available)

However it says as of setuptools 41.5.0, the test command is deprecated. It seems like "extras" is easier to specify?

PR #99, opened 2020-01-12
fix: support nested classes

### LINKED ISSUE #89: Argument conflict when trying to replace snapshottest
**Describe the bug**

Pytest raises an error from argparse because of a conflicting arguments: both snapshottest and syrupy declare a --snapshot-update argument, so you have to trash your virtualenv or explicitly uninstall snapshottest to get syrupy working.

**To Reproduce**

1. Create a project with snapshottest and pytest as a dependency (e.g. install it in a venv)
1. Install syrupy and remove snapshottest from your dependencies
1. Running pytest shows an error because of the duplicate argument
1. Delete your virutalenv and reinstall, or explicitly uninstall snapshottest
1. pytest works a-ok

**Expected behavior**

It would be nice if you didn't get this confusing error when migrating, it would make syrupy an easier drop-in replacement for snapshottest, but it seems like this may be out of control of plugins. Could be worth calling it out explicitly in the docs for other people who encounter it, if a programmatic solution is not available.

 - OS: MacOS
 - Version 10.14.5
 - Python 3.6.6, pip 18.1

### LINKED ISSUE #98: Support nested test classes
**Describe the bug**

Syrupy does not support nested test blocks

**To Reproduce**

*Given*

```py
# test_str.py
class TestClass:
    class TestSubClass:
        def test_sub_class_has_parent(self, snapshot):
            assert "this should be in a nested class" == snapshot
```

*Result*

```ambr
# name: TestSubClass.test_sub_class_has_parent
  'this should be in a nested class'
---
```

**Expected behavior**

```ambr
# name: TestClass.TestSubClass.test_sub_class_has_parent
  'this should be in a nested class'
---
```

**Screenshots**

N/A

 - Version 0.1.0

**Additional context**

N/A

iamogbz:
## Description

Refactors base extension to remove subdirectory prop.

Bumps the python package to beta.

## Related Issues

- Closes #89 
- Closes #98 

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-- Add support for nested test classes
-- Removes `_snapshot_subdirectory_name` from `SnapshotFossilizer`
+- Add support for nested test classes (#99)
+- Removes `_snapshot_subdirectory_name` from `SnapshotFossilizer` (#99)
 
 ## v0.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-Up to date with releases.
+- Removes `_snapshot_subdirectory_name` from `SnapshotFossilizer`
 
 ## v0.1.0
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-Up to date with releases.
+- Add support for nested test classes
```
noahnu: We should include the PR the feature/bugfix was added in. PR number is sufficient

PR #100, opened 2020-01-12
docs: custom object snapshot

### LINKED ISSUE #71: Provide better default repr for classes without overridden repr
**Describe the bug**

Snapshotting a class instance which does not explicitly define a serializable repr or str, will use the default object repr which includes the memory address of the class instance. This changes every test run.

**To Reproduce**

Add to a test file:

```
class MyClass:
    pass

def test_class(snapshot):
    assert snapshot = MyClass()
```

then

```
inv test -d -u
```

**Expected behavior**

Snapshots should not change between test runs if the code is deterministic.

**Additional context**

It's possible we don't want to try support this, as it clearly implies the data type is not meant to be serialized.

iamogbz:
## Description

Adds documentation for custom object snapshotting

## Related Issues

- Closes #71 

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,12 @@ def __repr__(self) -> str:
 ```
 
 It is recommended to split representation into multiple lines.
+
+https://github.com/tophat/syrupy/blob/9da5a72c153548870e5e9c2dd62fdd944ae7c663/tests/examples/test_custom_object_repr.py#L20-L24
+
 This makes the snapshot diff in case of failures or updates easy to review.
+
+https://github.com/tophat/syrupy/blob/9da5a72c153548870e5e9c2dd62fdd944ae7c663/tests/examples/__snapshots__/test_custom_object_repr.ambr#L5-L9
 
 ### Options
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -48,6 +48,37 @@ A snapshot file should be generated under a `__snapshots__` directory in the sam
...
+def __repr__(self) -> str:
+  return "MyCustomClass(...)"
+```
+
+It is recommended to split representation into multiple lines.
+
+```python
+def __repr__(self):
+  state = "\n".join(
+    f"  {a}={getattr(self, a)}" for a in dir(self) if not a.startswith("__")
```
noahnu: Should wrap dir in sorted in the example

tests/examples/test_custom_object_repr.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class MyCustomSmartReprClass:
+    prop1 = 1
+    prop2 = "a"
+    prop3 = {1, 2, 3}
+
+    def __repr__(self):
+        state = "\n".join(
+            f"  {a}={getattr(self, a)}" for a in dir(self) if not a.startswith("__")
```
noahnu: Sorted

PR #101, opened 2020-01-12
feat: default serialization for custom objects

iamogbz:
## Description

Adds a default serialization for custom objects

## Related Issues

N/A

## Checklist

## Additional Comments

Makes adoption easier

### DIFF
--- a/src/syrupy/extensions/amber.py
+++ b/src/syrupy/extensions/amber.py
@@ -192,7 +192,24 @@ class DataSerializer:
     def serialize_unknown(
         cls, data: Any, *, depth: int = 0, visited: Optional[Set[Any]] = None
     ) -> str:
-        return cls.with_indent(repr(data), depth)
+        data_repr = repr(data)
+        is_default_repr = object.__repr__(data) == data_repr
+        if not is_default_repr:
+            return cls.with_indent(data_repr, depth)
+
+        return (
+            cls.with_indent(f"{cls.object_type(data)} {{\n", depth)
+            + "".join(
+                f"{name}="
+                + cls.serialize(
+                    data=getattr(data, name), depth=depth + 1, visited=visited
+                ).lstrip(cls._indent)
+                + "\n"
+                for name in cls.sort(dir(data))
+                if not name.startswith("__")
+            )
+            + cls.with_indent("}", depth)
+        )
 
     @classmethod
     def serialize(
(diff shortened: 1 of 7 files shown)


tests/examples/__snapshots__/test_custom_object_repr.ambr
```diff
@@ -1,10 +1,18 @@
-# name: test_snapshot_attr_class
-  MyCustomReprClass(prop1=1, prop2='a', prop3={1, 2, 3})
+# name: test_snapshot_custom_class
+  <class 'MyCustomClass'> {
+  prop1=1
```
noahnu: We should indent the props

src/syrupy/extensions/amber.py
```diff
@@ -192,7 +192,26 @@ def serialize_iterable(
...
+                f"{serialized_key}={serialized_value.lstrip(cls._indent)}\n"
+                for serialized_key, serialized_value in (
+                    (
+                        cls.with_indent(name, depth=depth + 1),
+                        cls.serialize(
+                            data=getattr(data, name), depth=depth + 1, visited=visited
+                        ),
+                    )
+                    for name in cls.sort(dir(data))
+                    if not name.startswith("__")
```
noahnu: Thoughts on relaxing this to a single underscore?

iamogbz: Sure, it would prevent protected attributes from getting recorded

src/syrupy/extensions/amber.py
```diff
@@ -188,6 +188,18 @@ def serialize_iterable(
...
+    def serialize_callable(
+        cls,
+        data: "SerializableData",
+        *,
+        depth: int = 0,
+        visited: Optional[Set[Any]] = None,
+    ) -> str:
+        return cls.with_indent(
+            f"{cls.object_type(data)} ({cls.MarkerDepthMax()})", depth
+        )
```
noahnu: There are some options here. Should include the function name? Doc string? type information if available?

What's a use case where someone would snapshot a function?

iamogbz: We can definitely get the type information, but I was wondering if it's better to just ignore methods when in a class. That way we won't need to account for callable.

noahnu: Personally I'm in favour of ignoring and leaving this up to a community built extension of needed, or open issues.

PR #111, opened 2020-01-14
chore: pin pytest version to range

### LINKED ISSUE #110: Add pytest as a dependency
**Describe the bug**

Syrupy relies on pytest's APIs, specifically its hooks, and yet it doesn't specify what version of pytest is supported.

**Expected behavior**

Syrupy should list pytest in the setup.py with the maximum version range supported.

**Additional context**

syrupy v0.2.0

noahnu:
## Description

This PR specifies a version range for pytest.

Also went ahead and updated dev & test dependencies.

## Related Issues

- Closes #110.

## Checklist


setup.py
```diff
@@ -13,12 +13,12 @@
     "attrs>=18.2.0,<20.0.0",
     "colored>=1.3.92,<2.0.0",
     "typing_extensions>=3.6,<4.0.0",
+    "pytest>=5.0.0,<6.0.0",
 ]
 test_requires = [
     "codecov",
     "coverage[toml]",
     "invoke",
-    "pytest",
```
iamogbz: nb. should it be in both? As in both `install_requires` and `test_requires`

noahnu: we always install "install requires", so I don't think it's necessary

PR #114, opened 2020-01-15
fix: serialize carriage returns, #113

### LINKED ISSUE #113: Syrupy cannot handle carriage returns in strings
**Describe the bug**

If a carriage return is present in a string, Syrupy will strip out the carriage return when serializing the data thus making the assertion persistently fail even after a snapshot-update.

**To Reproduce**

```python
def test_carriage(snapshot):
    assert snapshot == "line 1\r\nline 2"
```

```shell
pytest --snapshot-update; pytest
```

**Expected behavior**

Syrupy should be able to serialize any control character.

**Additional context**

Syrupy v0.3.0

noahnu:
## Description

If a carriage return is present in a string, Syrupy will strip out the carriage return when serializing the data thus making the assertion persistently fail even after a snapshot-update.

This PR keeps the original newlines by explicitly setting the amber newlines arg of `open`.

## Related Issues

- Closes #113

## Checklist

## Additional Comments

This does not fix the assertion reporter. With this PR, if you remove the `\r` from one of the assertions in the added test, you will get an extremely unhelpful diff of "...".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-- Fix bug where newline control characters where being translated based on platform (#113)
+- Fix bug where newline control characters were being translated based on platform (#113)
 
 ## [v0.3.0](https://github.com/tophat/syrupy/compare/v0.2.0...v0.3.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-Up to date with releases.
+- Fix bug where newline control characters where being translated based on platform (#113)
 
 ## [v0.3.0](https://github.com/tophat/syrupy/compare/v0.2.0...v0.3.0)
 
--- a/src/syrupy/extensions/amber.py
+++ b/src/syrupy/extensions/amber.py
@@ -36,14 +36,15 @@ class DataSerializer:
         Writes the snapshot data into the snapshot file that be read later.
         """
         filepath = snapshot_fossil.location
-        with open(filepath, "w") as f:
+        with open(filepath, "w", newline="") as f:
             for snapshot in sorted(snapshot_fossil, key=lambda s: s.name):
                 snapshot_data = str(snapshot.data)
                 if snapshot_data is not None:
                     f.write(f"{cls._marker_name} {snapshot.name}\n")
-                    for data_line in snapshot_data.split("\n"):
-                        f.write(f"{cls._indent}{data_line}\n")
-                    f.write(f"{cls._marker_divider}\n")
+                    print(snapshot_data.splitlines(keepends=True))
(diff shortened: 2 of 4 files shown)


tests/test_extension_amber.py
```diff
@@ -17,6 +17,12 @@ def test_empty_snapshot(snapshot):
     assert snapshot == ""
 
 
+def test_newline_control_characters(snapshot):
+    assert snapshot == "line 1\nline 2"
+    assert snapshot == "line 1\r\nline 2"
```
noahnu: To see the broken reporting, remove the `\r`. I don't have a fix for the reporter.

src/syrupy/extensions/amber.py
```diff
@@ -56,12 +57,12 @@ def read_file(cls, filepath: str) -> "SnapshotFossil":
...
         snapshot_fossil = SnapshotFossil(location=filepath)
         try:
-            with open(filepath, "r") as f:
+            with open(filepath, "r", newline="") as f:
                 test_name = None
                 snapshot_data = ""
                 for line in f:
                     if line.startswith(cls._marker_name):
-                        test_name = line[name_marker_len:-1].strip(" \n")
+                        test_name = line[name_marker_len:-1].strip(" \r\n")
```
iamogbz: We should call out that amber does not preserve trailing newlines... nb. but then it's kinda obvious from the output and I doubt a lot of people would care

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
-Up to date with releases.
+- Fix bug where newline control characters where being translated based on platform (#113)
```
iamogbz:
```suggestion
- Fix bug where newline control characters were being translated based on platform (#113)
```

tests/test_extension_amber.py
iamogbz: Should add an issue to track this

src/syrupy/extensions/amber.py
noahnu: It does preserve trailing newlines for serialized content. This is for the test name.

tests/test_extension_amber.py
noahnu: https://github.com/tophat/syrupy/issues/116

src/syrupy/extensions/amber.py
```diff
@@ -36,14 +36,15 @@ def write_file(cls, snapshot_fossil: "SnapshotFossil") -> None:
...
-        with open(filepath, "w") as f:
+        with open(filepath, "w", newline="") as f:
             for snapshot in sorted(snapshot_fossil, key=lambda s: s.name):
                 snapshot_data = str(snapshot.data)
                 if snapshot_data is not None:
                     f.write(f"{cls._marker_name} {snapshot.name}\n")
-                    for data_line in snapshot_data.split("\n"):
-                        f.write(f"{cls._indent}{data_line}\n")
-                    f.write(f"{cls._marker_divider}\n")
+                    print(snapshot_data.splitlines(keepends=True))
```
iamogbz: Oops rogue print statement causing bugs

PR #127, opened 2020-02-09
Fix cleanup unused snapshots when targeting files

iamogbz:
## Description

- Fixes issue where snapshot files were not cleaned up when running specific test files
- Fixes issue where targeting specific test nodes in a test file was not supported
- Fixes issue where targeting specific test modules using pyargs was not supported

## Related Issues

N/A

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -65,7 +65,7 @@ def pytest_sessionstart(session: Any) -> None:
         update_snapshots=config.option.update_snapshots,
         base_dir=config.rootdir,
         is_providing_paths=any(
-            not arg.startswith("-") and glob.glob(arg)
+            not arg.startswith("-") and glob.glob(arg.split("::")[0])
             for arg in config.invocation_params.args
         ),
     )
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -54,6 +54,14 @@ def pytest_assertrepr_compare(op: str, left: Any, right: Any) -> Optional[List[s
     return None
 
 
+def __is_testpath(arg: str) -> bool:
+    return not arg.startswith("-") and bool(glob.glob(arg.split("::")[0]))
+
+
+def __is_testmodule(arg: str) -> bool:
+    return arg == "--pyargs"
+
+
 def pytest_sessionstart(session: Any) -> None:
     """
     Initialize snapshot session before tests are collected and ran.
@@ -65,7 +73,7 @@ def pytest_sessionstart(session: Any) -> None:
         update_snapshots=config.option.update_snapshots,
         base_dir=config.rootdir,
         is_providing_paths=any(
-            not arg.startswith("-") and glob.glob(arg.split("::")[0])
+            __is_testpath(arg) or __is_testmodule(arg)
             for arg in config.invocation_params.args
         ),
     )
--- a/src/syrupy/data.py
+++ b/src/syrupy/data.py
@@ -50,6 +50,8 @@ class SnapshotFossil(object):
 
     def add(self, snapshot: "Snapshot") -> None:
         self._snapshots[snapshot.name] = snapshot
+        if snapshot.name != SNAPSHOT_EMPTY_FOSSIL_KEY:
+            self.remove(SNAPSHOT_EMPTY_FOSSIL_KEY)
 
     def merge(self, snapshot_fossil: "SnapshotFossil") -> None:
         for snapshot in snapshot_fossil:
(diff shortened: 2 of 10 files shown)


src/syrupy/__init__.py
```diff
@@ -65,7 +65,7 @@ def pytest_sessionstart(session: Any) -> None:
         update_snapshots=config.option.update_snapshots,
         base_dir=config.rootdir,
         is_providing_paths=any(
-            not arg.startswith("-") and glob.glob(arg)
+            not arg.startswith("-") and glob.glob(arg.split("::")[0])
```
iamogbz: Support `pytest test_mod.py::TestClass::test_method`
https://docs.pytest.org/en/latest/usage.html

tests/test_integration_default.py
```diff
@@ -271,8 +319,8 @@ def test_unused_snapshots_warning(stubs):
     assert result.ret == 0
 
 
-def test_unused_snapshots_ignored_if_not_targeted_when_targeting_using_dash_k(stubs):
```
iamogbz: This was already covered by `test_collection` so repurposed to test for the node ids case

tests/test_integration_default.py
```diff
@@ -283,18 +331,18 @@ def test_unused_snapshots_ignored_if_not_targeted_when_targeting_using_dash_k(st
     testdir.makefile(
         ".py", **{"test_life_always_finds_a_way": test_content},
     )
+    testfile = os.path.join(testdir.tmpdir, "test_life_always_finds_a_way.py")
     result = testdir.runpytest(
-        "-k", "life_always_finds_a_way", "-v", "--snapshot-update"
```
iamogbz: This was already covered by `test_collection` so repurposed to test for the node ids case

src/syrupy/report.py
```diff
@@ -113,6 +98,12 @@ def unused(self) -> "SnapshotFossils":
             self.discovered, self.used
         ):
             snapshot_location = unused_snapshot_fossil.location
+            if self.is_providing_paths and not any(
+                TestLocation(node).matches_snapshot_location(snapshot_location)
```
iamogbz: This replaces the previous logic which was essentially
```py
if self.is_providing_paths and snapshot_location not in self.used:
    continue
```

src/syrupy/__init__.py
noahnu: hmm.. we should also check if we support the pyargs syntax: `--pyargs <module>`

src/syrupy/data.py
```diff
@@ -50,6 +50,8 @@ def get(self, snapshot_name: str) -> Optional["Snapshot"]:
 
     def add(self, snapshot: "Snapshot") -> None:
         self._snapshots[snapshot.name] = snapshot
+        if snapshot.name != SNAPSHOT_EMPTY_FOSSIL_KEY:
+            self.remove(SNAPSHOT_EMPTY_FOSSIL_KEY)
```
noahnu: What's the broken behaviour this specific change is fixing?

src/syrupy/extensions/base.py
```diff
@@ -123,6 +123,12 @@ def write_snapshot(self, *, data: "SerializedData", index: int) -> None:
         """
         self._pre_write(data=data, index=index)
         snapshot_location = self.get_location(index=index)
+        if not self.test_location.matches_snapshot_location(snapshot_location):
+            warning_msg = f"""
+            Can not relate snapshot location '{snapshot_location}' to the test location.
```
noahnu: (nb) i18n gettext

non-blocking since we still haven't done any sort of translations

tests/test_integration_default.py
```diff
@@ -5,7 +5,53 @@
 from .utils import clean_output
 
 
-def test_collection(testdir):
+def test_unused_snapshots_ignored_if_not_targeted_using_dash_m(testdir):
```
noahnu: interesting. Wasn't aware of `-m`.

src/syrupy/data.py
iamogbz: Nothing broken at the moment, but there should not be an empty snapshot fossil marker once snapshot have been added. Noticed it while attempting to fix the listed issue.

src/syrupy/extensions/base.py
```diff
@@ -124,17 +125,21 @@ def write_snapshot(self, *, data: "SerializedData", index: int) -> None:
         self._pre_write(data=data, index=index)
         snapshot_location = self.get_location(index=index)
         if not self.test_location.matches_snapshot_location(snapshot_location):
-            warning_msg = f"""
-            Can not relate snapshot location '{snapshot_location}' to the test location.
-            Consider adding '{self.test_location.filename}' to the generated location.
-            """
+            warning_msg = gettext(
+                """
+                Can not relate snapshot location '{}' to the test location.
```
noahnu: Unnecessary leading white space for this warning message and the one below.

PR #130, opened 2020-02-20
Replace all instances of `os.path` with `pathlib` calls

### LINKED ISSUE #121: Replacing `os.path` with `pathlib`
This is mainly a refactor concern: in Python 3.4+, [`pathlib`](https://docs.python.org/3/library/pathlib.html) can serve as a more powerful (and built-in) replacement for `os.path`, dealing with flexible pathlike objects compatible anywhere where a string path may be used. It offers extra features over just dealing with paths, such as navigating filetrees with ease using object methods. Succinctly put, it provides a cleaner experience when dealing with file structures.

Refactor `syrupy` so that it uses `pathlib` wherever it deals with paths instead of `os.path`.

Continuing with `os.path` and accepting increased complexity.

**Additional context**
Since Syrupy already requires `python >= 3.6`, the version requirement for `pathlib` isn't a concern.

mcataford:
## Description

As explained in #121, `pathlib` is the newer, py3 successor to `os.path` and provides an arguably friendlier interface to system paths than `os.path` does. Additionally, it is built to be platform agnostic, which is a hoot.

## Related Issues

This resolves #121 and replaces all instances of `os.path` with `pathlib.Path` methods. This encompasses both the source and the test code.

## Checklist

**Since the changes are entirely internal, this doesn't add any more tests (all present tests passing is sufficient to establish correctness and equivalency to the previous `os.path` patterns).**

- ~[ ] This PR has sufficient test coverage.~
- ~[ ] I have updated the CHANGELOG.md.~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ From v1.0.0 onwards, this project adheres to [Semantic Versioning](https://semve
 
 ## Master (Unreleased)
 
+- Conversion of all `os.path` and `os.walk` calls to use `pathlib` instead, setting `pathlib` as the new preferred way of doing path operations (#130)
+
 - Up to date with releases.
 
 ## [v0.3.2](https://github.com/tophat/syrupy/compare/v0.3.1...v0.3.2)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ These are mostly guidelines, not rules. Use your best judgment, and feel free to
 
 - [Commit Messages](#commit-messages)
 - [Code Styleguide](#code-styleguide)
+- [`pathlib` over `os.path`](#usage-of-pathlib)
 
 [Additional Notes](#additional-notes)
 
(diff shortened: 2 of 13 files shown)


mcataford:
@noahnu Since these changes are just internal refactoring, does it warrant a changelog entry?

mcataford:
@iamogbz That's a good question. Cursory search says no, but I can try and throw one together. 🤔

noahnu:
> @noahnu Since these changes are just internal refactoring, does it warrant a changelog entry?

No strong opinions on whether refactoring should be included in the changelog. @iamogbz thoughts?

There's no harm in including refactors in the changelog though. Perhaps we should retroactively prefix the messages with semantic commit prefixes to call out there's no functional change?

iamogbz:
> There's no harm in including refactors in the changelog though. Perhaps we should retroactively prefix the messages with semantic commit prefixes to call out there's no functional change?

@mcataford @noahnu no strong opinions either, but I think we should include this in the changelog, seeing as this is a code style change that we want to enforce going forward. Or maybe adding a blurb in the `CONTRIBUTING` doc would suffice.

mcataford:
Gotcha, in this case I'll add to the `CONTRIBUTING` re:the convention, and in the changelog, then reseek review. 👍

mcataford:
@noahnu @iamogbz Added some docs to changelog and contributing, see if it makes sense to you.

mcataford:
Yeet.

I'll leave y'all the action of merging to preserve the old chain of command.

PR #132, opened 2020-02-21
feat: Add --snapshot-default-extension option

### LINKED ISSUE #82: Configuration option to specify default serializer plugin

Currently you have to override the snapshot fixture in your project's root conftest.py in order to specify the default serializer class. It's potentially not clear what the final serializer is (order of shadowing).

A configuration option to specify the python module path to load for the default serializer.

```
--snapshot-default-plugin syrupy.extensions.image.SVGImageExtension
```

Adding documentation for overriding the default fixture.

msrose:
## Description

Add syrupy cli option to allow modifying default extension without adding a custom fixture

## Related Issues

- Closes #82 

## Checklist


src/syrupy/__init__.py
```diff
@@ -110,9 +119,18 @@ def pytest_sessionfinish(session: Any, exitstatus: int) -> None:
 
 @pytest.fixture
 def snapshot(request: Any) -> "SnapshotAssertion":
+    default_serializer_plugin = request.config.option.default_serializer_plugin
+
+    if default_serializer_plugin:
+        parts = default_serializer_plugin.split(".")
+        module, extension_class_name = ".".join(parts[:-1]), parts[-1:][0]
+        extension_class = getattr(import_module(module), extension_class_name)
```
msrose: Using importlib feels sort of hacky but I'm not sure of a different way. The import_module API is awkward since it requires the module to be extracted from the path passed to the command line. 

It might be best to move this to the command-line parsing phase and store the extension_class in request.config.option instead.

src/syrupy/__init__.py
```diff
@@ -38,6 +37,13 @@ def pytest_addoption(parser: Any) -> None:
         dest="warn_unused_snapshots",
         help="Do not fail on unused snapshots",
     )
+    group.addoption(
+        "--snapshot-default-plugin",
```
iamogbz: Wording should be `--snapshot-default-extension` since that's the language used elsewhere

src/syrupy/utils.py
```diff
@@ -14,3 +18,33 @@ def walk_snapshot_dir(root: str) -> Generator[str, None, None]:
...
+            f"Cannot load member '{module_member_name}' without module path"
+        )
+
+    try:
+        module = import_module(module_name)
+    except ModuleNotFoundError:
+        raise FailedToLoadModuleMember(f"Module '{module_name}' does not exist.")
+
+    try:
+        module_member = getattr(module, module_member_name)
```
iamogbz: Can't `import_module` directly import a single export in a module. Or is this to provide better error messages?

src/syrupy/utils.py
```diff
@@ -14,3 +18,33 @@ def walk_snapshot_dir(root: str) -> Generator[str, None, None]:
             continue
         if not filepath.name.startswith(".") and filepath.is_file():
             yield str(filepath)
+
+
+class FailedToLoadModuleMember(Exception):
```
iamogbz: There's a file for custom exceptions

src/syrupy/utils.py
```diff
@@ -14,3 +18,33 @@ def walk_snapshot_dir(root: str) -> Generator[str, None, None]:
...
+    pass
+
+
+def load_module_member_from_path(path: str) -> Any:
+    path_parts = path.split(".")
+
+    module_name = ".".join(path_parts[:-1])
+    module_member_name = path_parts[-1:][0]
```
iamogbz: Can be reduced to
```suggestion
    [*module_parts, module_member_name] = path.split(".")
    module_name = ".".join(module_parts)
```

src/syrupy/utils.py
iamogbz: Just tried, guess it can't and I can't see any clean way to do it

src/syrupy/__init__.py
iamogbz: > It might be best to move this to the command-line parsing phase and store the extension_class in request.config.option instead.

Could be a good idea, but not sure how it would be done or the benefit?

iamogbz: Just tried it out, seems to work quite well, [example](https://github.com/tophat/syrupy/compare/add-default-extension-option...iamogbz-patch-1)

src/syrupy/__init__.py
```diff
@@ -75,6 +92,7 @@ def pytest_sessionstart(session: Any) -> None:
     session._syrupy = SnapshotSession(
         warn_unused_snapshots=config.option.warn_unused_snapshots,
         update_snapshots=config.option.update_snapshots,
+        default_extension_class=config.option.default_extension,
```
noahnu: (nb) is there an advantage to storing the default in the session? We reference the request.config directly in the snapshot fixture on line 139.

src/syrupy/utils.py
```diff
@@ -14,3 +19,22 @@ def walk_snapshot_dir(root: str) -> Generator[str, None, None]:
             continue
         if not filepath.name.startswith(".") and filepath.is_file():
             yield str(filepath)
+
+
+def import_module_member(path: str) -> Any:
```
noahnu: `Any` or can we be specific and reference a concrete implementation of `AbstractSyrupyExtension` as we do here:  https://github.com/tophat/th_snapshot/blob/331c5d5e85310630bd55a101e3bd4744f51c770d/src/syrupy/assertion.py#L57

noahnu:
@all-contributors please add @msrose for code, tests

PR #160, opened 2020-03-21
fix: support call syntax for snapshot fixture overriding

iamogbz:
## Description

Support call syntax to easily switch extension class

## Related Issues

N/A

## Checklist

## Additional Comments

No additional comments.


iamogbz:
Using `fix:` inorder to trigger a patch release and not a minor version bump

tests/test_extension_image.py
```diff
@@ -8,40 +8,48 @@
...
-        '<g><g fill="#fff" stroke="#707070">'
-        '<rect width="50" height="50" stroke="none"/>'
-        '<rect x="0" y="0" width="50" height="50" fill="none"/></g>'
-        '<text transform="translate(10 27)" fill="#707070" '
-        'font-family="ConsolasForPowerline, Consolas for Powerline" font-size="8">'
-        '<tspan x="0" y="0">50 x 50</tspan></text></g></g></svg>'
-    )
-    assert snapshot_svg == actual_svg
+    assert actual_svg == snapshot(extension_class=SVGImageSnapshotExtension)
+    assert actual_svg == snapshot  # last extension class specified is used
```
noahnu: I feel like that behaviour can be confusing. Perhaps it should function similar to snapshot.use_extension, where it effectively clones the snapshot instance

iamogbz: If it clones it then it'll lose the context of assertion index, leading to duplicates

noahnu: we can clone assertion index. I find it strange that calling `snapshot` at the beginning of a test case can affect snapshot assertions towards the end of the test case. Strange side-effects

iamogbz: So your preferred behaviour would be having this line use the default snapshot extension class? It does make it more "pure" but then requires a lot of cloning or state values. Since the snapshot fixture is bound to the test case function the limit of the side effects would only be in specific test, I don't think that's too confusing and it would allow for doingt things like #161 without a lot of cloning.

iamogbz: @noahnu also I don't think we could track with cloning assertion index if each individual assertion in the test case is a new instance

tests/test_extension_image.py
```diff
@@ -8,40 +8,48 @@
...
-        '<g><g fill="#fff" stroke="#707070">'
-        '<rect width="50" height="50" stroke="none"/>'
-        '<rect x="0" y="0" width="50" height="50" fill="none"/></g>'
-        '<text transform="translate(10 27)" fill="#707070" '
-        'font-family="ConsolasForPowerline, Consolas for Powerline" font-size="8">'
-        '<tspan x="0" y="0">50 x 50</tspan></text></g></g></svg>'
-    )
-    assert snapshot_svg == actual_svg
+    assert actual_svg == snapshot(extension_class=SVGImageSnapshotExtension)
+    assert actual_svg == snapshot  # uses initial extension class
```
iamogbz: @noahnu changed, so that the initial extension class is used when not overridden

src/syrupy/assertion.py
```diff
@@ -70,6 +74,10 @@ def executions(self) -> Dict[int, AssertionResult]:
     def use_extension(
         self, extension_class: Optional[Type["AbstractSyrupyExtension"]] = None,
     ) -> "SnapshotAssertion":
+        """
+        Creates a new snapshot assertion fixture with the same options but using
+        specified extension class. This does not perserve assertion index or state.
```
noahnu:
```suggestion
        specified extension class. This does not preserve assertion index or state.
```

PR #172, opened 2020-04-03
Fix warnings

iamogbz:
## Description

- Fixes an issue where the assertion `_extension` property cache is always cleared
- Fixes warnings below
> PytestDeprecationWarning: TerminalReporter.writer attribute is deprecated, use TerminalReporter._tw instead at your own risk.
  See https://docs.pytest.org/en/latest/deprecations.html#terminalreporter-writer for more information.

## Related Issues

N/A

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -1,6 +1,7 @@
 from gettext import gettext
 from typing import (
     TYPE_CHECKING,
+    Callable,
     Dict,
     List,
     Optional,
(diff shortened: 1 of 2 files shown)


src/syrupy/assertion.py
```diff
@@ -44,10 +45,9 @@ class SnapshotAssertion:
     _test_location: "TestLocation" = attr.ib(kw_only=True)
     _update_snapshots: bool = attr.ib(kw_only=True)
     _extension: Optional["AbstractSyrupyExtension"] = attr.ib(init=False, default=None)
-    _executions: int = attr.ib(init=False, default=0, kw_only=True)
-    _execution_results: Dict[int, "AssertionResult"] = attr.ib(
-        init=False, factory=dict, kw_only=True
-    )
+    _executions: int = attr.ib(init=False, default=0)
+    _execution_results: Dict[int, "AssertionResult"] = attr.ib(init=False, factory=dict)
+    _post_assert_actions: List[Callable[..., None]] = attr.ib(init=False, factory=list)
```
iamogbz, line 50: Could use a `deque` but that seems like premature optimisations

tests/test_extension_image.py
```diff
@@ -51,5 +51,6 @@ def test_multiple_snapshot_extensions(snapshot):
     """
     assert actual_svg == snapshot(extension_class=SVGImageSnapshotExtension)
     assert actual_svg == snapshot  # uses initial extension class
+    assert snapshot._extension is not None
```
noahnu, line 54: This is testing internals. Is there a user facing test?

iamogbz: There's no user effect only that slight performance drawback of generating the extension instance on each assertion

noahnu: Bit concerned about coupling the internal implementation with the test cases, but am okay with this as a temporary measure until we have perf benchmarking

PR #177, opened 2020-04-08
fix: correctly track unused snapshots in classes

noahnu:
## Description

Fixes a bug where targeting a specific subset of tests in a test class would incorrectly report the non-targeted test cases as "unused".

## Checklist

### DIFF
--- a/tests/test_integration_default.py
+++ b/tests/test_integration_default.py
@@ -100,6 +100,55 @@ def test_unused_parameterized_ignored_if_not_targeted_using_dash_k(
     assert Path(*snapshot_path).joinpath("other_snapfile.ambr").exists()
 
 
+def test_only_updates_targeted_snapshot_in_class_for_single_file(testdir):
+    test_content = """
+        import pytest
+
+        class TestClass:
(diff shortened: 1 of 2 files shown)


tests/test_integration_default.py
```diff
@@ -100,6 +100,55 @@ def test_collected(snapshot, actual):
...
+    assert snapshot_path.joinpath("test_content.ambr").exists()
+
+    test_filepath = Path(testdir.tmpdir, "test_content.py")
+    result = testdir.runpytest(str(test_filepath), "-v", "-k test_case_2")
+    result_stdout = clean_output(result.stdout.str())
+    assert "1 snapshot passed" in result_stdout
+    assert "snapshot unused" not in result_stdout
+
+
+def test_only_updates_targeted_snapshot_for_single_file(testdir):
```
iamogbz, line 128: I think this was already covered in other tests

taion:
I think this leads to false negatives for subsequent snapshots; e.g. `__parse("foo.1")` returns `"foo.1"`, while it should return `"foo"`. Sending up a PR shortly.

PR #186, opened 2020-04-15
fix: Fix parsing identifiers from snapshot names

taion:
## Description

This fixes a regression introduced in #177 around generating parsed names from snapshot names with index suffixes.

## Related Issues

- #177

## Checklist

- [x] This PR has sufficient test coverage.
- [x] I will merge this pull request with a semantic title.

## Additional Comments

No additional comments.

PR #187, opened 2020-04-15
fix: parsing identifiers from snapshot names (#186)

noahnu:
See: https://github.com/tophat/syrupy/pull/186

Author: https://github.com/taion


noahnu:
@all-contributors please add @taion for code and tests

PR #192, opened 2020-04-17
feat: v0.4.0 - Amber serializer changes

### LINKED ISSUE #183: Multiline strings within dicts contain extra trailing whitespace in snapshots
**Describe the bug**

Multiline strings in a dictionary are serialized with extra trailing whitespace.

**To Reproduce**

test_file.py

```python
def test_lines(snapshot):
    lines = "\n".join([
        "line 1",
        "line 2",
        "line 3"
    ])
    assert { "value": lines } == snapshot
```

Output (whitespace highlighted):

image

**Expected behavior**

Trailing whitespace should only be included if it's actually present in the data that is serialized.

- Syrupy 0.3.9
- Python 3.8.1

**Additional context**

The amount of trailing whitespace seems to vary. Might be based on indentation, haven't looked into it too much.

This is consistent behaviour and thus does not "break" anything, and thus fixing this would require a breaking change.

### LINKED ISSUE #193: Multiline strings should be indented to the same depth as the parent

Multiline strings span multiple lines in the amber snapshot but are not indented. This leads to hard to read snapshots.

Use the same indentation rules as list/dict/other values. I.e. indent every line in the string.

noahnu:
## Description

v0.4.0 contains some breaking changes to the amber snapshot serializer.

## Related Issues

- Closes #183, removes added trailing whitespace from multiline string
- Closes #193, multiline strings should be indented according to parent

## Checklist

### DIFF
--- a/src/syrupy/extensions/amber.py
+++ b/src/syrupy/extensions/amber.py
@@ -103,7 +103,10 @@ class DataSerializer:
         if "\n" in data:
             return (
                 cls.with_indent("'\n", depth)
-                + str(data)
+                + "".join(
+                    cls.with_indent(line, depth + 1 if depth else depth)
+                    for line in str(data).splitlines(keepends=True)
+                )
                 + "\n"
                 + cls.with_indent("'", depth)
             )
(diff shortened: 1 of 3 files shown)


iamogbz:
@noahnu should probably be merged with a merge commit rather than a squash

PR #194, opened 2020-04-17
feat(amber): indent multiline strings, close #193

### LINKED ISSUE #193: Multiline strings should be indented to the same depth as the parent

noahnu:
## Description

Indents multiline strings only when the parent itself is indented, this is to keep snapshots of multiline strings that are not nested in any data structure, simple.

## Related Issues

- Closes #193, multiline strings should be indented according to parent

## Checklist

### DIFF
--- a/tests/__snapshots__/test_extension_amber.ambr
+++ b/tests/__snapshots__/test_extension_amber.ambr
@@ -76,7 +76,7 @@
     'd': ...,
   }
 ---
-# name: test_deeply_nested_list_in_dict
+# name: test_deeply_nested_multiline_string_in_dict
   <class 'dict'> {
     'value_a': <class 'dict'> {
       'value_b': '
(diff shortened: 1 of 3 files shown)


tests/__snapshots__/test_extension_amber.ambr
```diff
@@ -142,8 +153,8 @@
 # name: test_list_in_dict
```
iamogbz, line 153: This is weirdly named, more like `multi_line_in_dict`

tests/test_extension_amber.py
```diff
@@ -28,6 +28,12 @@ def test_list_in_dict(snapshot):
     assert {"value": lines} == snapshot
 
 
+def test_deeply_nested_list_in_dict(snapshot):
```
iamogbz, line 31: nb. technically not a list?
```suggestion
def test_deeply_nested_multi_line_in_dict(snapshot):
```

PR #197, opened 2020-04-19
fix: use the test node location when determining snapshot class name

### LINKED ISSUE #195: Shared test mixins that use the snapshot fixture save under wrong name
**Describe the bug**
Shared test mixins that use the snapshot fixture save under the mixin class name rather than the subclass name

**To Reproduce**
```py
class SharedMixin:
    def test_math(self, snapshot):
        assert 1 + 1 == snapshot

class TestMathWorks(SharedMixin):
    pass

class TestMathWorksAgain(SharedMixin):
    pass
```

**Expected behavior**
The snapshot is saved under the subclass names

iamogbz:
## Description

Use the pytest node location to parse the full class name path

## Related Issues

- Closes #195

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -1,6 +1,7 @@
 from pathlib import Path
 from typing import (
     Any,
+    Iterator,
     Optional,
 )
 
@@ -16,8 +17,8 @@ class TestLocation:
 
     @property
     def classname(self) -> Optional[str]:
-        classes = self._node.obj.__qualname__.split(".")[:-1]
-        return ".".join(classes) if classes else None
+        _, __, qualname = self._node.location
+        return ".".join(list(self.__valid_ids(qualname))[:-1]) or None
 
     @property
     def filename(self) -> str:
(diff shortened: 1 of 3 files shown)


src/syrupy/location.py
```diff
@@ -16,8 +17,8 @@ def __init__(self, node: Any):
 
     @property
     def classname(self) -> Optional[str]:
-        classes = self._node.obj.__qualname__.split(".")[:-1]
-        return ".".join(classes) if classes else None
+        _, __, qualname = self._node.location
```
noahnu, line 20: Where are you seeing documentation for "location"?

noahnu: ah found it. This is "Item" not "Node": https://docs.pytest.org/en/latest/_modules/_pytest/nodes.html#Item

src/syrupy/location.py
```diff
@@ -16,8 +17,8 @@ def __init__(self, node: Any):
...
+        return ".".join(list(self.__valid_ids(qualname))[:-1]) or None
```
lime-green, line 21: the `list()` is probably unnecessary

src/syrupy/location.py
noahnu: Traced the source code:

https://github.com/pytest-dev/pytest/blob/7d5f5a878598aaa7f3e100ef6a97bcd13c1b408d/src/_pytest/python.py#L328 -> https://github.com/pytest-dev/pytest/blob/7d5f5a878598aaa7f3e100ef6a97bcd13c1b408d/src/_pytest/python.py#L296

iamogbz: I think it's `Function` which inherits from `Item` and for our use `PyobjMixin`
https://docs.pytest.org/en/latest/_modules/_pytest/python.html#Function

With the `location` property from `Node` using the `reportinfo` method defined in `PyobjMixin`

https://github.com/pytest-dev/pytest/blob/7d5f5a878598aaa7f3e100ef6a97bcd13c1b408d/src/_pytest/python.py#L315-L330

PR #200, opened 2020-04-21
fix: Handle dotted parameters in classname

taion:
## Description

This fixes a regression from https://github.com/tophat/syrupy/pull/197 for tests with parameters with dots in them.

## Related Issues

https://github.com/tophat/syrupy/pull/197, sorta

## Checklist

- [ ] I will merge this pull request with a semantic title.

## Additional Comments

Without this change, the dotted parameter name would confuse the earlier iteration of the `classname` logic.

### DIFF
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -5,6 +5,8 @@ from typing import (
     Optional,
 )
 
+from pytest import Class
+
 
 class TestLocation:
     def __init__(self, node: Any):
@@ -17,8 +19,12 @@ class TestLocation:
 
     @property
     def classname(self) -> Optional[str]:
-        _, __, qualname = self._node.location
-        return ".".join(list(self.__valid_ids(qualname))[:-1]) or None
+        return (
+            ".".join(
+                node.name for node in self._node.listchain() if isinstance(node, Class)
+            )
+            or None
+        )
 
     @property
     def filename(self) -> str:
--- a/tests/test_extension_amber.py
+++ b/tests/test_extension_amber.py
@@ -169,3 +169,15 @@ class TestClass:
 
 class TestSubClass(TestClass):
     pass
+
+
+@pytest.mark.parametrize("parameter_with_dot", ("value.with.dot",))
+def test_parameter_with_dot(parameter_with_dot, snapshot):
+    assert parameter_with_dot == snapshot
+
+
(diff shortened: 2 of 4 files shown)


tests/test_extension_amber.py
```diff
@@ -169,3 +169,15 @@ def test_nested_class_method(self, snapshot, actual):
...
+
+
+@pytest.mark.parametrize("parameter_with_dot", ("value.with.dot",))
+def test_parameter_with_dot(parameter_with_dot, snapshot):
+    assert parameter_with_dot == snapshot
+
+
+@pytest.mark.parametrize("parameter_1", ("foo",))
+@pytest.mark.parametrize("parameter_2", ("bar",))
+def test_doubly_parametrized(parameter_1, parameter_2, snapshot):
```
taion, line 181: I was about to implement something that just split on `[`, on closer inspection, it seemed better to traverse the node's parents.

src/syrupy/location.py
```diff
@@ -17,8 +19,12 @@ def __init__(self, node: Any):
...
     @property
     def classname(self) -> Optional[str]:
-        _, __, qualname = self._node.location
-        return ".".join(list(self.__valid_ids(qualname))[:-1]) or None
+        return (
+            ".".join(
+                node.name for node in self._node.listchain() if isinstance(node, Class)
+            )
+            or None
+        )
```
iamogbz, line 27: Would prefer we fixed the `__valid_ids` method so both `__parse` and `classname` are using the same logic

taion: I suspect `classname` doesn't need that logic – unlike with the others, in this context, we know that we have some set of valid Python identifiers anyway. It seems like it might, then, be slightly better to actually look for classes in the node hierarchy, as per this implementation?

iamogbz: Hmm missed this yesterday, it's an easy enough refactor to revert. @noahnu what do you think?

taion: Right, I mean, that's what this implementation does – `listchain` shows all the current "nodes" in the test chain (modules, classes, &c.), so I'm going off that directly.

The alternative is just to split on `[` and ignoring the part after that in `__valid_ids` before splitting on `.`, which gets you to the same place.

PR #239, opened 2020-05-27
Update setup.py, so that it allows for any Python 3.6+ version to be used

arturbalabanov:
## Description

The project supports any Python 3.6+ version, so we need to reflect this in `setup.py`.

## Related Issues

- Closes #238: Python version in setup.py is outdated

## Checklist

- [x] This PR has sufficient test coverage.
- [x] I will merge this pull request with a semantic title.

## Additional Comments

No additional comments.

PR #240, opened 2020-05-27
Update setup.py, so that it allows for any Python 3.6+ version to be used

noahnu:
PR: https://github.com/tophat/syrupy/pull/239


noahnu:
@all-contributors add arturbalabanov for code

Issue #149, opened 2020-03-04
Support ignoring fields when serializing


iamogbz:
**Is your feature request related to a problem? Please describe.**

When asserting on snapshots support filtering out properties that should not be matched on. This would allow deterministic snapshots when matching on objects that generate realtime values in tests without needing to mock.

**Describe the solution you'd like**

```py
def test_my_dict(snapshot):
	my_dict = {
		"match_me": True,
		"do_not_match_me": time.time(),
		"nested_do_not_match": {
			"do_not_match_me": time.time(),
		},
	}
	assert my_dict == snapshot(exclude=("do_not_match_me",))
```
```ambr
# name: test_my_dict
  <class 'dict'> {
    'match_me': True,
	'nested_do_not_match': <class 'dict'> {
	},
  }
---
```

**Describe alternatives you've considered**

This is already possible but is cumbersome to implement, and cannot easily be extended to other serialization types.

```py
class OmitDataSerializer(DataSerializer):
	@classmethod
	def __filter_data(cls, data) -> Dict[str, Any]:
		return {
			key: data[key]
			for key in data
			if key not in ("do_not_match_me",)
		}
	
    @classmethod
    def serialize_dict(cls, data, **kwargs) -> str:
        return super().serialize_dict(cls.__filter_data(data), **kwargs)


class OmitDataExtension(AmberSnapshotExtension):
    def serialize(self, data):
        return OmitDataSerializer.serialize(data)


def test_my_dict(snapshot):
	my_dict = {
		"match_me": True,
		"do_not_match_me": time.time(),
		"nested_do_not_match": {
			"do_not_match_me": time.time(),
		},
	}
	assert my_dict == snapshot(extension=OmitDataExtension)
```

**Additional context**

- Supported by Jest https://jestjs.io/docs/en/snapshot-testing#property-matchers
- Much requested feature, see counterpart https://github.com/syrusakbary/snapshottest/issues/21

noahnu:
Thoughts on `include` and `exclude` instead of `ignore`?

We could support strings (literal property matchers) and predicates

noahnu:
> We could replace the bool return with Optional[Any] and only exclude if it's None otherwise replace with the new value

That feels more like a "transformation" function. A `bool` should be sufficient, and if we support some sort of transformation, it would be in some other parameter (not exclude).

taion:
We use this sort of approach internally. I took a stab at porting what I had over into something more Syrupy-native, _but_ for us there's a twist.

[...]

PR #245, opened 2020-05-31
feat(amber): add property matcher support

iamogbz:
## Description

Support property matchers
- Also makes arguments optional when using call syntax
- Adds helper matcher factory for ease of adoption
https://github.com/tophat/syrupy/blob/6fa24059755cc29524ce223993bd3e6392ecb6d5/tests/test_matchers.py#L28-L33

## Related Issues

- Relates to #149, a short description of the linked issue.

## Checklist

## Additional Comments

- Jest https://jestjs.io/docs/en/snapshot-testing#property-matchers
- Snapshottest syrusakbary/snapshottest#21


src/syrupy/types.py
```diff
@@ -1,8 +1,14 @@
...
+    Hashable,
+    Optional,
+    Tuple,
     Union,
 )
 
 
 SerializableData = Any
 SerializedData = Union[str, bytes]
+PropertyPath = Tuple[Hashable, ...]
```
iamogbz, line 13: Should types be included in the path sequence? i.e.
```suggestion
PropertyPath = Tuple[Tuple[Hashable, Type[SerializableData]], ...]
```

src/syrupy/extensions/amber.py
```diff
@@ -145,17 +158,21 @@ def serialize_dict(
...
     ) -> str:
-        kwargs = {"depth": depth + 1, "visited": visited}
+        kwargs = {"depth": depth + 1, "matcher": matcher, "visited": visited}
         return (
             cls.with_indent(f"{cls.object_type(data)} {{\n", depth)
             + "".join(
                 f"{serialized_key}: {serialized_value.lstrip(cls._indent)},\n"
                 for serialized_key, serialized_value in (
                     (
                         cls.serialize(**{"data": key, **kwargs}),
```
iamogbz, line 172: Opted to not include the existing path in the serialization of object keys even though the matcher is included

src/syrupy/assertion.py
```diff
@@ -88,10 +89,13 @@ def use_extension(
     def assert_match(self, data: "SerializableData") -> None:
         assert self == data
 
+    def _serialize(self, data: "SerializableData") -> "SerializedData":
+        return self.extension.serialize(data, matcher=self._matcher)
```
iamogbz, line 93: Might be a breaking change if extension serializers were not written with support for variable keyword arguments i.e. `**kwargs`.

src/syrupy/types.py
iamogbz: Including types would allow for matching on the parent types

src/syrupy/assertion.py
noahnu: Worth forcing it with *?

src/syrupy/extensions/amber.py
```diff
@@ -242,16 +289,21 @@ def serialize(
...
+        path: "PropertyPath" = (),
         visited: Optional[Set[Any]] = None,
     ) -> str:
         visited = visited if visited is not None else set()
         data_id = id(data)
         if depth > cls._max_depth or data_id in visited:
             data = cls.MarkerDepthMax()
-
+        if matcher:
+            data = matcher(data, path)
```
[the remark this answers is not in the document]
iamogbz: I used the term "matcher" because it behaves almost exactly as jest's [asymmetric matcher](https://jestjs.io/docs/en/snapshot-testing#property-matchers)

iamogbz:
> Generally LGTM. Can you update the README.md? We don't have this new behaviour documented anywhere

@noahnu will do. Do you have any thoughts on the potential breaking nature of the change.

noahnu:
Since the serializer is a public API, yes I believe it's a breaking change. How do we prevent this in the future? A note to use **kwargs? What do other python projects do?

src/syrupy/types.py
iamogbz: @noahnu any thoughts on this?

src/syrupy/assertion.py
iamogbz: > Since the serializer is a public API, yes I believe it's a breaking change. How do we prevent this in the future? A note to use **kwargs? What do other python projects do?

@noahnu we could force the use of kwargs in the serializer?

```py
def serialize(self, *, **kwargs)...
```

src/syrupy/extensions/amber.py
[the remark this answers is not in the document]
noahnu: It looks like it behaves a bit differently. In this case, the "matcher" is a transformation function which takes some data and a path and returns a value to serialize in place of the original data. Something closer would be:

```python
def test_case(snapshot):
    assert { "a": { "b": 1, "c": "uuid" } } == snapshot(matcher=(("a","b"),))
```

```python
if matcher:
   if any(path.startswith(prefix) for prefix in matcher):
       data = Excluded(data) # special helper class for non-deterministic data
```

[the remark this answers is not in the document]
iamogbz: I don't think it behaves in the exclusion manner, since the jest snapshots are being replaced with values like `Any<Number>` and `Any<Date>`. The difference being with jest you have limited control over what the replacement snapshot looks like.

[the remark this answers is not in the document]
iamogbz: @noahnu see example snapshot showing usage
https://github.com/tophat/syrupy/blob/06ff63786ae63ba40d391b7e076d76a4044fc1cc/tests/__snapshots__/test_extension_amber.ambr#L217-L227

README.md
```diff
@@ -80,7 +80,74 @@ These are the cli options exposed to `pytest` by the plugin.
...
+
+**NOTE:** Do not mutate the value received as it could cause unintended side effects.
+
+##### Built-In Matchers
+
+Syrupy comes with a few built-in matcher presets that can be used to make easy work of using property matchers.
+
+###### `path_type(mapping=None, *, types=(), strict=True)`
+
+Easy way to build a matcher that uses the path and value type to replace serialized.
```
noahnu, line 108:
```suggestion
Easy way to build a matcher that uses the path and value type to replace serialized data.
```

README.md
```diff
@@ -80,7 +80,74 @@ These are the cli options exposed to `pytest` by the plugin.
...
+When strict, this will raise a `ValueError` if the types specified are not matched.
+
+| Argument  | Description                                          |
+| --------- | ---------------------------------------------------- |
+| `mapping` | Dict of path string to class type tuples             |
+| `types`   | The types to match if no mapped path does            |
```
noahnu, line 114: > The types to match if no mapped path does

Wording is a bit confusing. Can you also provide an example of "types"?

noahnu:
If you recall the Django case that came up, where we don't want to load an attribute, is there a way to match before the attribute is accessed? Some way to check type without triggering a property get.

src/syrupy/extensions/amber/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+        snapshot = self._read_snapshot_fossil(snapshot_location).get(snapshot_name)
+        return snapshot.data if snapshot else None
+
+    def _write_snapshot_fossil(self, *, snapshot_fossil: "SnapshotFossil") -> None:
+        snapshot_fossil_to_update = DataSerializer.read_file(snapshot_fossil.location)
+        snapshot_fossil_to_update.merge(snapshot_fossil)
+        DataSerializer.write_file(snapshot_fossil_to_update)
+
+
+__all__ = ["AmberSnapshotExtension", "DataSerializer"]
```
iamogbz, line 71: @noahnu keeping this for backwards compat, but I think the `DataSerializer` should be an internal interface

README.md
```diff
@@ -80,7 +80,74 @@ These are the cli options exposed to `pytest` by the plugin.
...
+| Argument | Description                                                                   |
+| -------- | ----------------------------------------------------------------------------- |
+| `data`   | Current serializable value being matched on                                   |
+| `path`   | Ordered path traversed to the current value e.g. `(("a", dict), ("b", dict))` |
+
+**NOTE:** Do not mutate the value received as it could cause unintended side effects.
+
+##### Built-In Matchers
+
+Syrupy comes with a few built-in matcher presets that can be used to make easy work of using property matchers.
```
noahnu, line 104: You mention a "few". Are they all documented here? Path_type is more of a matcher factory

README.md
```diff
@@ -80,7 +80,74 @@ These are the cli options exposed to `pytest` by the plugin.
...
+
+###### `path_type(mapping=None, *, types=(), strict=True)`
+
+Easy way to build a matcher that uses the path and value type to replace serialized data.
+When strict, this will raise a `ValueError` if the types specified are not matched.
+
+| Argument  | Description                                                                                                                        |
+| --------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `mapping` | Dict of path string to tuples of class types                                                                                       |
+| `types`   | Tuple of class types used if none of the path strings from the mapping are matched                                                 |
```
noahnu, line 114: Does it only work on class types? What about primitive types?

iamogbz: Primitives work as well e.g. https://github.com/tophat/syrupy/blob/6fa24059755cc29524ce223993bd3e6392ecb6d5/tests/test_matchers.py#L28-L33

README.md
iamogbz: There's only one so far, will update the wording

PR #247, opened 2020-06-01
fix: only process valid test nodes in report, close #246

### LINKED ISSUE #246: Syrupy crashes when used with pytest-flake8
## Bug description

With the following dependencies in my poetry file:

```py
syrupy = "^0.4.3"
pytest = "^5.4.2"
pytest-flake8 = "^1.0.6"
```

I have this test file:

test_snapshot.py
```py
def test_a_snapshot(snapshot):
    assert 1 == snapshot
```

And when I run it I get:

```
$ py.test --flake8 test_snapshot.py 
============================= test session starts =============================
... blah ...                            

test_snapshot.py::FLAKE8 PASSED                                         [ 50%]
test_snapshot.py::test_a_snapshot PASSED                                [100%]
Traceback (most recent call last):
...
  File "/path/to/venv/lib/python3.7/site-packages/syrupy/report.py", line 104, in <genexpr>
    for node in self.ran_items
  File "/path/to/venv/lib/python3.7/site-packages/syrupy/location.py", line 13, in __init__
    self.modulename = self._node.obj.__module__
AttributeError: 'Flake8Item' object has no attribute 'obj'
```

## To Reproduce
1. Install pytest-flake8 and syrupy
2. Write a test in a test file that uses snapshot
3. run `py.test --flake8 path/to/your/test.py`

## Expected behaviour
The test should not fail in tear down and py.test should exit with 0 exit code.

## Screenshots
N/A

## Environment
 - OS: macOS 10.15.3 Catalina
 - Python Version 3.7.4

## Extra info

FWIW I "fixed" this for my install by using the following monkey-patch:

```py
# Monkey patch Flake8Item due to a conflict with syrupy https://github.com/tophat/syrupy/issues/246
import pytest_flake8
pytest_flake8.Flake8Item.obj = type('DummyFlake8Obj', (), {})
```

But I think it's really a problem with syrupy for assuming the `obj` attribute will be on the pytest `Item`.

noahnu:
## Description

pytest-flake8 seems to create nodes not associated with test cases such that `obj` is not defined on the Flake8 Item. It seems that Flake8 adds nodes to the test collection without running the test itself, so snapshot is never called, which means this is only a concern in test item collection. A check for the obj property may be sufficient, although a better solution might be some way to identify a "real" test?

## Related Issues

- Closes #246, syrupy crashes when used with pytest-flake8

## Checklist


src/syrupy/__init__.py
```diff
@@ -114,15 +114,17 @@ def pytest_collection_modifyitems(session: Any, config: Any, items: List[Any]) -
     After tests are collected and before any modification is performed.
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_modifyitems
     """
-    config._syrupy._all_items.update(items)
+    valid_items = (item for item in session.items if hasattr(item, "obj"))
```
iamogbz, line 117: Shouldn't this be `items`
```suggestion
    valid_items = (item for item in items if hasattr(item, "obj"))
```

src/syrupy/__init__.py
```diff
@@ -114,15 +114,17 @@ def pytest_collection_modifyitems(session: Any, config: Any, items: List[Any]) -
...
+    config._syrupy._all_items.update(valid_items)
 
 
 def pytest_collection_finish(session: Any) -> None:
     """
     After collection has been performed and modified.
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_finish
     """
-    session.config._syrupy._ran_items.update(session.items)
+    valid_items = (item for item in session.items if hasattr(item, "obj"))
```
iamogbz, line 126: We can separate this out into a function to reuse

noahnu:
I dived into the Flake8 and pytest source.

A Flake8Item inherits from File and Item, https://github.com/tholo/pytest-flake8/blob/e2461d12d82523db55f4d156d1c060da1e66b21e/pytest_flake8.py#L95:

```python
class Flake8Item(pytest.Item, pytest.File):
```

From pytest, it looks like only Function Items have an obj: https://github.com/pytest-dev/pytest/blob/694fdc655436b6b2eb335d9019002ab81486e4d8/src/_pytest/python.py#L1420. An "Item" is an individual test invocation.

It seems we've conflated the concept of "node" with "item" in some cases, although in the report collector it is in fact an "Item". An FSCollector, when provided with a path, should return a "Module" (or Package). A "Module" is itself a PyCollector which collects all items in the module, creating items or collectors. The idea being that a Module's collect method may return a leaf node such as a Function(Item), or a non-leaf node such as a Class which itself would be a collector.

There is an example in pytest of a File which is also an Item: https://github.com/pytest-dev/pytest/blob/5034399d7acc4bd14fdad3d056a9abc2fde13863/testing/example_scripts/fixtures/custom_item/conftest.py#L4. So it seems there is official support by pytest of this use case. It's definitely strange though, since without any plugins/flake8 extensions, a File is usually a collector and would thus never end up in the collected items.

I believe the better fix would be to check that the Item is an instance of a Function node. Is there a use case where a snapshot would be asserted outside of a Function node?

noahnu:
Having trouble writing a test for this. I've tried:

```
from .utils import clean_output
def test_ignores_non_function_nodes(testdir):
    conftest = (
        """
        import pytest
        class CustomItem(pytest.Item, pytest.File):
            def __init__(self, *args, **kwargs):
                super().__init__(*args, **kwargs)
                self._nodeid += "::CUSTOM"
            def runtest(self):
                pass
        def pytest_collect_file(path, parent):
            return CustomItem(path, parent)
        """
    )
    testcase = (
        """
        def test_example(snapshot):
            assert snapshot == 1
        """
    )
    testdir.makepyfile(conftest=conftest)
    testdir.makepyfile(test_file=testcase)
    result = testdir.runpytest("-v", "--snapshot-update")
    result_stdout = clean_output(result.stdout.str())
    assert result.ret == 0
    assert "test_file.py::CUSTOM" in result_stdout
```

however it always passed.

It also seems the flake8 issue only happens 1 out of every 3 times, even with cache disabled. The behaviour alternates.

noahnu:
I can reproduce using:

```
       import pytest

        class CustomItem(pytest.Item, pytest.File):
            def __init__(self, *args, **kwargs):
                super().__init__(*args, **kwargs)
                self._nodeid += "::CUSTOM"

            def runtest(self):
                pass

        def pytest_collect_file(path, parent):
            return CustomItem(path, parent)
```

without pytest-flake8. Unfortunately it's still flakey and only happens sometimes.

src/syrupy/report.py
```diff
@@ -36,8 +36,8 @@
 @attr.s
 class SnapshotReport:
     base_dir: str = attr.ib()
-    all_items: Set[Any] = attr.ib()
-    ran_items: Set[Any] = attr.ib()
+    all_items: Dict["pytest.Item", bool] = attr.ib()
```
noahnu, line 39: changed this to a Dict to ensure iteration in deterministic order. Shouldn't affect performance, and the deterministic order allows us to properly test this bug fix

PR #377, opened 2020-10-14
feat(amber): normalise line endings between os

### LINKED ISSUE #199: Test syrupy in windows environment

An issue was discovered with windows compatibility. Testing on windows would have caught the issue.

Run ci tests in windows environment

N/A

**Additional context**
- #196, issue raised limited to windows os
- [Pillow example](https://github.com/nulano/Pillow/blob/master/.github/workflows/test-windows.yml)
- Github actions [windows-latest](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idruns-on)

iamogbz:
## Description

- [x] Run tests on windows

## Related Issues

- Closes #199

## Checklist

## Additional Comments

Exposing the windows runner for debugging

```yml
  debug:
    name: Debug
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@v1
      - name: Set up Python 3.8
        uses: actions/setup-python@v1
        with:
          python-version: 3.8
      - name: Start SSH Session
        uses: luchihoratiu/debug-via-ssh@v1.0
        with:
          NGROK_AUTH_TOKEN: ${{ secrets.DEBUG_NGROK_TOKEN }}
          SSH_PASS: ${{ secrets.DEBUG_SSH_PASSWORD }}
```


src/syrupy/extensions/amber/serializer.py
```diff
@@ -73,20 +73,21 @@ def read_file(cls, filepath: str) -> "SnapshotFossil":
...
+                        test_name = line[name_marker_len:-1].strip(f" {crn}")
                         snapshot_data = ""
                         continue
                     elif test_name is not None:
                         if line.startswith(cls._indent):
                             snapshot_data += line[indent_len:]
                         elif line.startswith(cls._marker_divider) and snapshot_data:
                             snapshot_fossil.add(
-                                Snapshot(name=test_name, data=snapshot_data[:-1])
+                                Snapshot(name=test_name, data=snapshot_data.rstrip(crn))
```
noahnu, line 90: This is not going to work / it won't be accurate.

Previously we only stripped the last character, which we _knew_ was a `\n` that we added and thus not part of the snapshot data. With `rstrip` it strips an arbitrary amount of carriage returns and newlines. It will even strip real data, if someone were to snapshot a newline at the end of their string.

We have full control over what we write in the snapshots, so I would stick with the "\n". No need to write carriage returns in snapshots for windows. Windows won't add carriage returns unless we tell it to.

tasks.py
```diff
@@ -29,10 +34,10 @@ def requirements(ctx, upgrade=False):
     ]
     if upgrade:
         args.append("--upgrade")
-    ctx.run(
+    ctx_run(
+        ctx,
         f"echo '-e .[dev]' | python -m piptools compile "
         f"{' '.join(args)} - -qo- | sed '/^-e / d' > dev_requirements.txt",
```
noahnu, line 40: How does the windows runner work? Does it have bash?

src/syrupy/extensions/amber/serializer.py
iamogbz: For amber this will be just before a new section is added to the snapshot, i.e. before the `---` which is terminated with a closing tag before the new line we add that's why this still works.

iamogbz: @noahnu windows seems to behave differently when reading files that have newlines, at least I suspect that due the failing test logs. It would be better to have a consistent way of indicating in amber files that the snapshot data is ended and anything afterwards is a system new line and that can be applied and stipped regardless of the os newline characters.

tasks.py
iamogbz: Unsure but from the [logs](https://pipelines.actions.githubusercontent.com/mJ2awWhNnBYs06r7hgESwBnBIrvKkXMWwIl1fyScZT5thXu53o/_apis/pipelines/1/runs/2334/signedlogcontent/10?urlExpires=2020-10-15T22%3A31%3A48.9489027Z&urlSigningMethod=HMACV1&urlSignature=zrT5wJ8UIFrCOs5jDVYduv7qOFryWLet%2BwDfZ%2BUBX0s%3D) it seems to use `C:\Program Files\PowerShell\7\pwsh.EXE`

src/syrupy/extensions/amber/serializer.py
noahnu: It shouldn't behave any differently since python handles the OS for us. I suspect something else is going on.

> Note Python doesn’t depend on the underlying operating system’s notion of text files; all the processing is done by Python itself, and is therefore platform-independent.
>
> &mdash; https://docs.python.org/3/library/functions.html#open

We specify how newlines are too be processed:

```
with open(filepath, "r", encoding="utf-8", newline="")
```

When `newline=""` we use universal line mode, so we read the exact same content on Windows / Mac / Linux.

iamogbz: I would think the same thing but the logs seem to contradict and in the documentation (a little further down)
> If it is '', universal newlines mode is enabled, but line endings are returned to the caller untranslated
So it might be possible that the file on windows when read on windows has both `\r` and `\n` and because we use `""` not `None` the line endings are not translated into `\n` on input?

src/syrupy/extensions/amber/serializer.py
```diff
@@ -95,6 +99,23 @@ def read_file(cls, filepath: str) -> "SnapshotFossil":
...
+        data: "SerializableData",
+        *,
+        exclude: Optional["PropertyFilter"] = None,
+        matcher: Optional["PropertyMatcher"] = None,
+    ) -> str:
+        """
+        After serializing, new line control characters are normalised. This is needed
+        for interoperablity of snapshot matching between systems that do not use the
+        same new line control characters. Example snapshots generated on windows os
+        should not break when running the tests on a unix based system and vice versa.
```
iamogbz, line 112: Worth calling out as effectively snapshots will never show a control character mismatch.

.github/workflows/pythonapp.yml
```diff
@@ -8,33 +8,34 @@ env:
...
   tests:
     name: Tests
-    runs-on: ubuntu-18.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        os: [ubuntu-latest, windows-latest]
         python-version: [3.6, 3.7, 3.8]
-      fail-fast: true
+      fail-fast: false
```
noahnu, line 31: Why change the `fail-fast` argument?

src/syrupy/__init__.py
```diff
@@ -87,15 +87,16 @@ def pytest_assertrepr_compare(
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_assertrepr_compare
     """
     with __terminal_color(config):
+        recieved_name = received_style("[+ received]")
```
noahnu, line 90:
```suggestion
        received_name = received_style("[+ received]")
```

src/syrupy/__init__.py
```diff
@@ -87,15 +87,16 @@ def pytest_assertrepr_compare(
...
-            assert_msg = reset(
-                f"{snapshot_style(left.name)} {op} {received_style('received')}"
-            )
+            assert_msg = reset(f"{snapshot_name(left.name)} {op} {recieved_name}")
             return [assert_msg] + left.get_assert_diff()
         elif isinstance(right, SnapshotAssertion):
-            assert_msg = reset(
-                f"{received_style('received')} {op} {snapshot_style(right.name)}"
-            )
+            assert_msg = reset(f"{recieved_name} {op} {snapshot_name(right.name)}")
```
noahnu, line 99:
```suggestion
            assert_msg = reset(f"{received_name} {op} {snapshot_name(right.name)}")
```

src/syrupy/__init__.py
```diff
@@ -87,15 +87,16 @@ def pytest_assertrepr_compare(
...
+        recieved_name = received_style("[+ received]")
+
+        def snapshot_name(name: str) -> str:
+            return snapshot_style(f"[- {name}]")
+
         if isinstance(left, SnapshotAssertion):
-            assert_msg = reset(
-                f"{snapshot_style(left.name)} {op} {received_style('received')}"
-            )
+            assert_msg = reset(f"{snapshot_name(left.name)} {op} {recieved_name}")
```
noahnu, line 96:
```suggestion
            assert_msg = reset(f"{snapshot_name(left.name)} {op} {received_name}")
```

src/syrupy/__init__.py
```diff
@@ -87,15 +87,16 @@ def pytest_assertrepr_compare(
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_assertrepr_compare
     """
     with __terminal_color(config):
+        recieved_name = received_style("[+ received]")
+
+        def snapshot_name(name: str) -> str:
+            return snapshot_style(f"[- {name}]")
```
noahnu, line 93: Opportunistic formatting change?

src/syrupy/extensions/amber/serializer.py
```diff
@@ -73,20 +74,23 @@ def read_file(cls, filepath: str) -> "SnapshotFossil":
...
                         continue
                     elif test_name is not None:
                         if line.startswith(cls._indent):
                             snapshot_data += line[indent_len:]
                         elif line.startswith(cls._marker_divider) and snapshot_data:
                             snapshot_fossil.add(
-                                Snapshot(name=test_name, data=snapshot_data[:-1])
+                                Snapshot(
+                                    name=test_name,
+                                    data=snapshot_data.rstrip(os.linesep),
```
noahnu, line 92: Consider the snapshot generated via:

```
assert snapshot == "apple\n"
```

Is there concern that we might strip newlines that are actually part of the data? Do we need to limit the rstrip to the length of the linesep? If there is concern, let's add a test for it (although I thought we did have one?).

src/syrupy/extensions/amber/serializer.py
noahnu: I'm okay with this, however it's a breaking change.

src/syrupy/extensions/amber/serializer.py
```diff
@@ -95,6 +99,23 @@ def read_file(cls, filepath: str) -> "SnapshotFossil":
...
+        matcher: Optional["PropertyMatcher"] = None,
+    ) -> str:
+        """
+        After serializing, new line control characters are normalised. This is needed
+        for interoperablity of snapshot matching between systems that do not use the
+        same new line control characters. Example snapshots generated on windows os
+        should not break when running the tests on a unix based system and vice versa.
+        """
+        serialized = cls._serialize(data, exclude=exclude, matcher=matcher)
+        return serialized.replace("\r\n", "\n").replace("\r", "\n")
```
noahnu, line 115: I think I'm having trouble reasoning around how this impacts intentional carriage returns in a snapshot. Are we saying users are just not able to snapshot carriage returns at all? No way to separate an intentional carriage return from one coming from the OS?

.github/workflows/pythonapp.yml
iamogbz: More information to see if the failures are OS/version specific

src/syrupy/__init__.py
iamogbz: Yup, easier to see which diff line is related to what, both for accessibility and when color output is disabled

src/syrupy/extensions/amber/serializer.py
iamogbz: We already have tests that should cover it, basically the snapshot that gets generated is something like
```ambr
  '
  apple
  
  '
```
The last line with the `'` is what gets `rstrip`'d and that not part of the snapshot content to assert.

src/syrupy/extensions/amber/serializer.py
iamogbz: All carriage returns are translated into python newlines `"\n"`. Basically if you have `"\r\r\n"` it should end up being `"\n\n"` across all platforms.

src/syrupy/extensions/amber/serializer.py
iamogbz: Technically it should not be, in the sense that bumping to a new version will not break existing tests. Will add a test with single carriage returns to the before code to test this out

noahnu:
Going to test this out on an internal repo.

src/syrupy/extensions/amber/serializer.py
iamogbz: Now I'm thinking since it's a breaking change how about we roll our own line end character for amber files; i.e. `\r` and `\r` get translated into ambr specific values and then terminated with ambr specific line end character; e.g.
```py
assert "a\rb\r\nc\nd" == snapshot
```
```ambr
  '
  a␍
  b␍␤
  c␤
  '
```

noahnu: We'd then switch to byte mote for reading/writing the snapshots?

Carriage return vs. Newline doesn't happen too frequently, and I find it does make it a bit harder to read the raw snapshot file that way (mainly due to the introduction of "noise")

iamogbz: This would be a wider sweeping change though, than currently where only a few files need be updated

noahnu (approved):
LGTM

Make sure to mark as a breaking change. Might also be worth adding a note to the readme explaining how line endings are handled?

PR #394, opened 2020-10-27
fix: unused snapshot detection for targeting single parameterized test case

### LINKED ISSUE #137: Clarify test methods for integration tests and testing in general
Current tests are starting to be hard to reason about. Some clarification about the behaviour each test is attempting to verify is getting more essential.

Make each test case group a single file testing all the behaviours with module and test doc string explaining the purpose

N/A

**Additional context**
The [`tests/test_integration_default.py`](https://github.com/tophat/syrupy/blob/3e46a842ddfde309e46830918d380da5da2c04b8/tests/test_integration_default.py) file is an example of this complexity.

### LINKED ISSUE #333: Incorrect unused snapshot detection for targeting single case in parameterized test
**To Reproduce**

- Create a parameterized test case, for example test_dict in tests/test_extension_amber.py.
- Run pytest targeting one case: `pytest tests/test_extension_amber.py::test_dict[actual4]`

Syrupy says 1 snapshot passed, and the rest are unused.

**Expected behavior**

One snapshot should pass. Nothing should be mentioned about the non-targeted test cases.

---

Syrupy==0.6.1

iamogbz:
## Description

Fix unused snapshot detection for targeting single parameterized test case.
Clarify test methods for integration tests and testing in general.

Additionally
- Fix issue where snapshot location can be `__snapshot__/test_file.` where extension is an empty string.

## Related Issues

- Closes #137, module testing strategy
- Closes #333, snapshot usage detection

## Checklist

## Additional Comments

No additional comments.


src/syrupy/report.py
```diff
@@ -98,23 +143,25 @@ def unused(self) -> "SnapshotFossils":
...
             else:
                 unused_snapshots = {
                     snapshot
                     for snapshot in unused_snapshot_fossil
-                    if any(
-                        TestLocation(node).matches_snapshot_name(snapshot.name)
-                        for node in self.ran_items
-                    )
+                    if self._selected_items_match_name(snapshot.name)
+                    or self._provided_nodes_match_name(snapshot.name, provided_nodes)
```
iamogbz, line 164: This is the main gist of the fix. Basically `_selected_items_match_name` use the keyword expressions when they are present or matches against the ran items when no keywords provided or `_provided_nodes_match_name` checks if the nodes given for the test file matching this fossil location matches the snapshot name.

iamogbz, line 164: This is the main gist of the fix. Basically `_selected_items_match_name` use the keyword expressions when they are present or matches against the ran items when no keywords provided or `_provided_nodes_match_name` checks if the nodes given for the test file matching this fossil location matches the snapshot name.

src/syrupy/report.py
```diff
@@ -89,7 +134,7 @@ def num_unused(self) -> int:
 
     @property
     def ran_all_collected_tests(self) -> bool:
-        return self.all_items == self.ran_items and not self.is_providing_nodes
+        return self.all_items == self.ran_items
```
iamogbz, line 137: Moved the ` and not self.is_providing_nodes` check to the point of usage at `if self.ran_all_collected_tests and not any(provided_nodes):`, makes the method name more accurate.

src/syrupy/report.py
```diff
@@ -67,6 +75,43 @@ def __attrs_post_init__(self) -> None:
...
+                elif maybe_opt_arg is None:
+                    maybe_opt_arg = arg
+                    continue  # do not reset maybe_opt_arg
+            else:
+                arg_groups.append((maybe_opt_arg, arg.strip()))
+
+            maybe_opt_arg = None
+
+        for maybe_opt_arg, arg_value in arg_groups:
+            if maybe_opt_arg == "-k":  # or maybe_opt_arg == "-m":
```
iamogbz, line 98: Ignoring the marker expression as they can not be used to match against snapshot names and are already used to select which pytest items in the collection to run.

src/syrupy/location.py
```diff
@@ -5,8 +5,10 @@
     Optional,
 )
 
+from syrupy.constants import PYTEST_NODE_SEP
 
-class TestLocation:
+
+class PyTestLocation:
```
iamogbz, line 11: Changed the name because pytest was complaining when attempting to collect the class as a test.

src/syrupy/report.py
```diff
@@ -8,18 +8,22 @@
...
     Iterator,
     List,
+    Optional,
+    Set,
+    Tuple,
 )
 
 import attr
-import pytest
+from _pytest.mark.expression import Expression
```
iamogbz, line 17: Alternative to importing the protected class is to roll our own parsing of `-k` expressions.

tests/integration/test_snapshot_option_update.py
```diff
@@ -0,0 +1,411 @@
...
+            r"1 snapshot passed\. 1 snapshot updated\. 1 unused snapshot deleted\.",
+            r"Deleted test_used\[3\] \(__snapshots__[\\/]test_used.ambr\)",
+        )
+    )
+    snapshot_path = [testdir.tmpdir, "__snapshots__"]
+    assert Path(*snapshot_path, "test_used.ambr").exists()
+    assert Path(*snapshot_path, "test_updated_1.ambr").exists()
+
+
+def test_update_targets_only_selected_parametrized_tests_for_update_dash_k(
```
iamogbz, line 218: This test was updated to catch the case raised by this issue

src/syrupy/report.py
noahnu: Just confirming this works in the min. version of pytest we support?

src/syrupy/report.py
```diff
@@ -67,6 +75,43 @@ def __attrs_post_init__(self) -> None:
...
+            else:
+                arg_groups.append((maybe_opt_arg, arg.strip()))
+
+            maybe_opt_arg = None
+
+        for maybe_opt_arg, arg_value in arg_groups:
+            if maybe_opt_arg == "-k":  # or maybe_opt_arg == "-m":
+                self._keyword_expressions.add(Expression.compile(arg_value))
+            elif maybe_opt_arg is None:
+                import importlib
```
noahnu, line 101: Any reason to constraint the import to this scope?

iamogbz: Nope, oversight will update

src/syrupy/report.py
```diff
@@ -67,6 +75,43 @@ def __attrs_post_init__(self) -> None:
                 else:
                     self.failed.update(snapshot_fossil)
 
+    def __parse_invocation_args(self) -> None:
```
noahnu, line 78: Complexity has grown quite a bit. I'm wondering if we should do a pass afterwards and ensure sufficient docstrings to make contributing to syrupy easier

src/syrupy/report.py
iamogbz: Haven't will test that out, probably should have a way to automate the CI range test

iamogbz: Yeah no this does not work with pytest 5...

tests/integration/test_snapshot_option_defaults.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.fixture
+def run_testcases(testdir, testcases):
+    pyfile_content = "\n\n".join(testcases.values())
+    testdir.makepyfile(test_file=pyfile_content)
+    return testdir.runpytest("-v")
+
+
+def test_injected_fixture(run_testcases):
```
noahnu, line 23: I don't understand. What are we testing here? Just that the fixture works? Isn't that technically covered by other tests? Or is this just a more explicit test?

src/syrupy/report.py
iamogbz: Makes sense

iamogbz: Added

src/syrupy/report.py
iamogbz: Changed this to include a naive version of the `Expression` class for compatibility with older pytest versions while still giving us some of the scaffolding for keyword snapshot inclusion

noahnu (approved):
LGTM

PR #397, opened 2020-10-29
fix: support python 3.9

noahnu:


dev_requirements.txt
```diff
@@ -9,13 +9,15 @@ attrs==20.2.0             # via flake8-bugbear, pytest, syrupy
...
+cffi==1.14.3              # via cryptography
 chardet==3.0.4            # via requests
 click==7.1.2              # via black, pip-tools
-codecov==2.1.10            # via syrupy
+codecov==2.1.10           # via syrupy
 colorama==0.4.4           # via twine
 colored==1.4.2            # via syrupy
 configparser==5.0.1       # via py-githooks
 coverage[toml]==5.3       # via codecov, syrupy
+cryptography==3.2.1       # via secretstorage
```
noahnu, line 20: This comes from secretstorage which is a direct dependency of keyring. It's only required on linux and so disappears whenever requirements are updated manually on Mac.

setup.py
```diff
@@ -11,7 +11,7 @@
 install_requires = [
     "attrs>=18.2.0,<21.0.0",
     "colored>=1.3.92,<2.0.0",
-    "typing_extensions>=3.6,<4.0.0",
+    "typing_extensions>=3.6,<4.0.0; python_version < '3.7.0'",
```
noahnu, line 14: No need to install typing_extensions on >=3.7.

PR #398, opened 2020-10-29
fix: expand single filename legal characters

### LINKED ISSUE #400: Expand single filename default allowed character set
### Description

Currently we support a limited character set for the default filenames in the single file extension.

Expanding the list would allow snapshot to use all the characters available to the file system without needing to override the internal method.

e.g.

`test_parametrized_file[a].png` instead of `test_parametrized_filea.png`

Source:
https://github.com/tophat/syrupy/pull/398#discussion_r514699905_

iamogbz:
## Description

Closes #400; Allow any character in a snapshot file name except
- `\s` gets replaced with `_`
- `\/?%*:|"<>,;=` get dropped
- Control characters (`C`) get dropped

## Checklist

## Additional Comments

- List of reserved filename characters: [wiki/Filename/reserved](https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words)
- Unicode character categories [unicode/category/table](http://www.unicode.org/reports/tr44/#GC_Values_Table)
- Unicode characters by category [unicode/category/chars](https://www.compart.com/en/unicode/category)


src/syrupy/extensions/single_file.py
```diff
@@ -70,4 +70,4 @@ def _write_snapshot_fossil(self, *, snapshot_fossil: "SnapshotFossil") -> None:
     def __clean_filename(self, filename: str) -> str:
         filename = str(filename).strip().replace(" ", "_")
         max_filename_length = 255 - len(self._file_extension or "")
-        return re.sub(r"(?u)[^-\w.]", "", filename)[:max_filename_length]
+        return re.sub(r"(?u)[^-\w.\[\]]", "", filename)[:max_filename_length]
```
iamogbz, line 73: Likely a breaking change, since snapshot files would get renamed

iamogbz: If we are going to do this then it's probably worth supporting the full range of characters which user can narrow down if they choose to override

noahnu: What's the overlap between characters that can appear in a test name and characters supported by all three operating systems we support?

noahnu (dismissed):
Makes sense to me. It is a breaking change though

src/syrupy/extensions/single_file.py
iamogbz: I'm thinking we set the list as any characters except `\/:*"<>|` `NUL` and control characters.

noahnu: I do wonder if a whitelist is more reliable.

Also if we expand the character set, arguably we shouldn't replace whitespace with underscores, since whitespace is allowed on Windows/Mac/Linux. Personally, I like the "simpler" characters b/c they're easier to work with in the terminal. Something like whitesapce requires you to escape the char. There's a trade off here I suppose.

iamogbz: I too like the simple charset. I'm inclined to leave it as is for now until there's pressure to support more characters by default... no need to introduce a breaking change at the moment. I can just create an issue to track this, we should probably resolve it before v1 though

src/syrupy/extensions/single_file.py
```diff
@@ -69,5 +69,12 @@ def _write_snapshot_fossil(self, *, snapshot_fossil: "SnapshotFossil") -> None:
...
-        return re.sub(r"(?u)[^-\w.]", "", filename)[:max_filename_length]
+        exclude_chars = '\\/:*"<>|?'
+        exclude_categ = ("C",)
+        cleaned_filename = "".join(
+            c
+            for c in filename
+            if c not in exclude_chars
+            and not any(categ in category(c) for categ in exclude_categ)
+        )
+        return cleaned_filename[:255]  # maximum filename length
```
noahnu, line 80: Why remove the `- len(self._file_extension)`?

iamogbz: Oops reverted

noahnu (approved):

PR #399, opened 2020-10-30
build: test min install requires versions

noahnu:
## Description

Refactors CI to perform one test run using the minimum supported dependencies.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -44,8 +44,7 @@ jobs:
       - name: Test
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: |
-          invoke test --coverage
+        run: invoke test --coverage
   version_checks:
     name: Dependency Version Constraint Checks
     runs-on: ubuntu-latest
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -38,12 +38,13 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install project dependencies
-        run: . script/bootstrap
+        run: |
+          python -m pip install -U pip --no-cache-dir
+          python -m pip install -e . -r dev_requirements.txt
       - name: Test
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: |
-          . $VENV/bin/activate
           invoke test --coverage
   version_checks:
     name: Dependency Version Constraint Checks
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - uses: .github/actions/detect-python
+      - uses: ./.github/actions/detect-python
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -68,7 +68,7 @@ def readme() -> str:
         packages=find_packages("./src"),
         zip_safe=False,
         entry_points={"pytest11": ["syrupy = syrupy"]},
-        extras_require={"dev": dev_requires},
+        extras_require={"dev": dev_requires, "test": test_requires},
```
iamogbz, line 71: I remember this was removed before, just not sure why now

.github/workflows/pythonapp.yml
```diff
@@ -43,11 +44,32 @@ jobs:
...
+        run: |
+          invoke test --coverage
+  version_checks:
+    name: Dependency Version Constraint Checks
+    runs-on: ubuntu-latest
+    env:
+      SKIP_DEPS: 1
+    steps:
+      - uses: actions/checkout@v1
+      - uses: ./.github/actions/detect-python
```
iamogbz, line 56: Does this work without the leading `./`

setup.py
noahnu: We had no need for it I think.

script/bootstrap
```diff
@@ -18,6 +15,16 @@ else
     fi
 fi
 
+VENV="${VENV:-syrupy.venv}"
+test -d $VENV || python3 -m venv $VENV || return
+. $VENV/bin/activate
```
iamogbz, line 20: Do we need to use the venv in CI?

noahnu: Originally added in case I needed to uninstall all requirements. Technically don't need it now, but it does make it easier to work with dependencies. It's an extra line wherever we execute a script, so I don't see any harm in it.

.github/workflows/pythonapp.yml
noahnu: Just tested it. You need the `./`.

script/bootstrap
iamogbz: Personally prefer not having the extra line, makes it cleaner imo (nit pick though)

noahnu: It's also testing the bootstrap script for local dev in a sense. Don't feel too strongly about it.

.github/workflows/pythonapp.yml
```diff
@@ -43,11 +44,32 @@ jobs:
       - name: Test
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: invoke test --coverage
+        run: |
+          invoke test --coverage
```
iamogbz, line 48: Missed replacing the `python -m pip install` for this step?

tests/test_integration_pytest_extension.py
```diff
@@ -1,3 +1,9 @@
+import pytest
+
+pytest_version = tuple(int(v) for v in pytest.__version__.split("."))
+
+
+@pytest.mark.skipif(pytest_version < (6, 0), reason="at least pytest 6 required")
```
iamogbz, line 6: What part of this test is pytest 6 dependent? Is there a way to widen the support?

iamogbz (approved):
LGTM

.github/workflows/pythonapp.yml
noahnu: I'll remove the newline. Can't replace python -m pip install b/c of windows

tests/test_integration_pytest_extension.py
noahnu: `from_parent` doesn't exist on pytest.Item

noahnu: I actually recall changing this when upgrading to pytest 6. The older method fails in pytest 6, part of the suite of deprecations

PR #404, opened 2020-10-30
build: do not install local version of syrupy in dev env

iamogbz:
## Description

This allows the local test to be run against the current code rather than the installed version after boostrapping

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -25,7 +25,7 @@ if [[ -z $SKIP_DEPS ]]; then
     if [[ -z $CI ]]; then
         export PYTHONPATH="./src:$PYTHONPATH"
     else
-        python -m pip install -e .
+        python -m pip install -e . --no-deps
     fi
     python -m pip install -r dev_requirements.txt
 fi
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -22,7 +22,12 @@ test -d $VENV || python3 -m venv $VENV || return
 python -m pip install -U pip --no-cache-dir
 
 if [[ -z $SKIP_DEPS ]]; then
-    python -m pip install -e . -r dev_requirements.txt
+    if [[ -z $CI ]]; then
+        export PYTHONPATH="./src:$PYTHONPATH"
+    else
+        python -m pip install -e .
+    fi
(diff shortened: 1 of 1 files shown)


script/bootstrap
```diff
@@ -22,7 +22,12 @@ test -d $VENV || python3 -m venv $VENV || return
 python -m pip install -U pip --no-cache-dir
 
 if [[ -z $SKIP_DEPS ]]; then
-    python -m pip install -e . -r dev_requirements.txt
+    if [[ -z $CI ]]; then
+        export PYTHONPATH="./src:$PYTHONPATH"
+    else
+        python -m pip install -e .
```
noahnu, line 28: dev_requirements.txt includes the install_requires. You can change this to `install -e . --no-deps`

iamogbz:
```suggestion
        python -m pip install -e . --no-deps
```

script/bootstrap
```diff
@@ -22,7 +22,12 @@ test -d $VENV || python3 -m venv $VENV || return
 python -m pip install -U pip --no-cache-dir
 
 if [[ -z $SKIP_DEPS ]]; then
-    python -m pip install -e . -r dev_requirements.txt
+    if [[ -z $CI ]]; then
+        export PYTHONPATH="./src:$PYTHONPATH"
+    else
+        python -m pip install -e . --no-deps
```
noahnu, line 28: What's the rational for not installing syrupy and using `inv test -d` in local dev?

I was under the impression the `-e` ensures `inv test` will use the source code.

iamogbz: Hmm looking at the flags yeah it should. I was running into an where my changes were not reflected in the test run after bootstrap, will double check if I can reproduce.

iamogbz: I think I just missed this flag when running the install command manually after recreating my venv. Ran it manually since I can't source the bootstrap script in fish shell 🐟 . Tried repro in bash and zsh but it worked as expected, could run `pytest` without needing to export `PYTHONPATH`, same with fish after using `bash ./script/bootstrap`. Will close 👍

iamogbz:
None issue.

PR #435, opened 2020-11-27
chore: run ci on pull request target, close #434

### LINKED ISSUE #434: Run CI on forked pull requests

Currently CI doesn't run on pull requests from forked repos and so we have to manually setup a temp. branch.

Use related pull request action triggers.

noahnu:
## Description

I believe this should run CI for pull requests based with branches in forked repos.

## Related Issues

- Closes #434

## Checklist

## Additional Comments

https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories-2

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -1,7 +1,7 @@
 name: Syrupy CICD
 
 on:
-  - pull_request_target
+  - pull_request
 
 env:
   CI: 1
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -1,7 +1,6 @@
 name: Syrupy CICD
 
 on:
-  - push
   - pull_request_target
 
 env:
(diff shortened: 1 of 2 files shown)


noahnu:
This may cause duplicate actions.. I wonder if there's a way to have it only run one or the other? Will close this for now.

iamogbz:
> This may cause duplicate actions

Is this confirmed? If so then we could just pick the one that runs on both?

noahnu:
> > This may cause duplicate actions
> 
> Is this confirmed? If so then we could just pick the one that runs on both?

Not confirmed but it's what I believe based on some discussions I read. We should be able to test this fairly easily. I'd probably just create a new github project to test that out. If they do create duplicate actions, we'd need to go with pull_request_target rather than push.

.github/workflows/cicd.yml
```diff
@@ -1,6 +1,8 @@
 name: Syrupy CICD
 
-on: [push]
+on:
+  - push
+  - pull_request_target
```
noahnu, line 5: Confirmed this will trigger the job twice:

image

PR #440, opened 2020-12-01
feat: add PEP-561 py.typed file for downstream type checkers

### LINKED ISSUE #439: Mark as PEP 561 (type checking) compatible

MyPy and other static type checkers are helpful for validating code. Currently syrupy doesn't work perfectly in projects that use mypy, because it is not considered PEP 561 compatible: https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages

This means that something like:

```python
# syrupy-typing.py
from syrupy.assertion import SnapshotAssertion

def test_foo(snapshot: SnapshotAssertion) -> None:
    assert "foo" == snapshot(match="x")
```

fails with

```
syrupy-typing.py:2: error: Skipping analyzing 'syrupy': found module but no type hints or library stubs
    from syrupy import SnapshotAssertion
    ^
syrupy-typing.py:2: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

The fix for users for this is ignoring the imports (as described in the link there), such as `from syrupy.assertion import SnapshotAssertion # type: ignore`. However, this is unfortunate because it stops mypy and potentially other tooling from picking up the error: `match` instead of `matcher`.

By adding a `py.typed` file as described in https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages , the existing type annotations become available to mypy users.

The output of the above then becomes:

```
syrupy-typing.py:5: error: Unexpected keyword argument "match" for "__call__" of "SnapshotAssertion"; did you mean "matcher"?
        assert "foo" == snapshot(match="x")
                        ^
```

A separate stub file/project seems unnecessary, given there's already type annotations inline in the code.

**Additional context**

N/A

huonw:
## Description

This follows PEP-561 to mark syrupy as compatible with type
checking (driven by the type annotations that already exist and are
checked), by creating an empty `py.typed` in the project. This stops
users from having to ignore syrupy when type checking, and allows the
type checker to provide specific guidance about possible problems.

See https://mypy.readthedocs.io/en/latest/installed_packages.html#making-pep-561-compatible-packages for more information about this approach.

## Related Issues

- Closes #439, discussing the value for type checking

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,7 @@ if __name__ in ["__main__", "builtins"]:
             "write_to": "version.txt",
         },
         package_dir={"": "src"},
+        package_data={"": ["py.typed"]},
         packages=find_packages("./src"),
         zip_safe=False,
         entry_points={"pytest11": ["syrupy = syrupy"]},


noahnu:
Thank you for opening a PR, and welcome to syrupy!

I don't believe the `py.typed` file is being packaged with syrupy. You can download a dev version of this PR via:

```shell
pip install -i https://test.pypi.org/simple/ syrupy==2020.12.1.132857705147
```

You'll need to update the setup.py to include this additional package data, see:
- setup.py: https://github.com/tophat/syrupy/blob/620fb4a07e2aaaeec22bf9209826f7ef3ad7f014/setup.py#L50
- package_data: https://docs.python.org/3/distutils/setupscript.html#installing-package-data

huonw:
Thanks, and good catch. I'd missed that part of the mypy docs about `py.typed`, and only tried it with a `pip install -e .`'d version (which of course works locally since it's not moving any files around)!

I've now added a `package_data` similar to `package_dir`, and validated that it works locally with `pip install .` (i.e. no `-e`): the original version omits `py.typed` in the final `site-packages` directory but the latest commit includes it.

noahnu (approved):
Confirmed package_data worked by installing via:

```shell
pip install -i https://test.pypi.org/simple/ syrupy==2020.12.1.231511194070
```

noahnu:
@all-contributors please add huonw for code

huonw:
Thanks for merging so promptly and even doing a release! We're now using 1.1 in our code and it works great.

PR #505, opened 2021-05-30
feat(amber): change serialization to be py syntax like

### LINKED ISSUE #502: How to improve a large snapshot instance and make it more readable
I parsed an XML file and obtained a very large instance which contain many objects and lists. I want to make the parsed instance more readable and testable with snapshot data.

Here are my questions:

- What's the best practice to make a snapshot of a large instance?
- How to make the snapshot file more readable (code highlighting, line breaks)? 
- My IDE (VSCode) doesn't recognize `.ambr` file, and will not have code highlighting automatically for that snapshot file. Any solution?

## Snapshot Approach 1: stringify `self.__dict__`

```python
def __repr__(self) -> str:
    return f"{type(self).__name__}({self.__dict__})"
```

output:

```
  Context({xml_id': 1234, 'links': [...], ...})  # a very long line
```

Problem: 

The class `Context` has many list properties and the line of snapshot would be very long (> 10,000 columns). It's hard to read and compare.

## Snapshot Approach 2: JSON serialization

```python
def __repr__(self) -> str:
    return json.dumps(self.__dict__, indent=2)
```

output:

```
  {
    "xml_id": 1234, 
    "links": [
      ...
    ], 
    ...
  }
```

Problem: 

It's hard to add types to custom objects like the format `Context({ ... })` with json format.

Any suggestion?

iamogbz:
## Description

* This changes the default amber serialization to better match python syntax. This does **NOT** need the file to be python parseable, just leverages python syntax recognition to make snapshots easier to grok.
* Enforces keyword arguments in filter and matcher factory functions

### Example

* VSCode Python Language Mode
```json
    "files.associations": {
        "*.ambr": "python",
    },
```

https://github.com/tophat/syrupy/blob/aa46082894e3bb646d789a6f9c09c059d80dcd18/tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr#L49-L83

## Related Issues

- Closes #502, improve snapshot readability.

## Checklist

## Additional Comments

**BREAKING CHANGE**: update to serialization requires regeneration of snapshots

### Migration Guide
* `pytest --snapshot-update` to regenerate amber snapshots


src/syrupy/extensions/amber/serializer.py
```diff
@@ -61,7 +61,7 @@ def write_file(cls, snapshot_fossil: "SnapshotFossil") -> None:
                 if snapshot_data is not None:
                     f.write(f"{cls._marker_name} {snapshot.name}\n")
                     for data_line in snapshot_data.splitlines(keepends=True):
-                        f.write(cls.with_indent(data_line, 1))
+                        f.write(cls.with_indent(data_line, 0))
```
iamogbz, line 64: None snapshot lines should be clearly indicated with a `marker` so this indent is not needed

src/syrupy/extensions/amber/serializer.py
```diff
@@ -45,7 +45,7 @@ def __repr__(self) -> str:
 class DataSerializer:
     _indent: str = "  "
     _max_depth: int = 99
-    _marker_divider: str = "---"
+    _marker_divider: str = "# ---"
```
iamogbz, line 48: Changed to match python comment syntax

src/syrupy/extensions/amber/serializer.py
```diff
@@ -175,8 +174,8 @@ def serialize_string(
                 for line in str(data).splitlines(keepends=True)
             ),
             depth=depth,
-            open_tag="'",
-            close_tag="'",
+            open_tag="'''",
+            close_tag="'''",
```
iamogbz, line 178: Python syntax multiline strings

src/syrupy/extensions/amber/serializer.py
```diff
@@ -266,7 +265,7 @@ def sort(cls, iterable: Iterable[Any]) -> Iterable[Any]:
 
     @classmethod
     def object_type(cls, data: "SerializableData") -> str:
-        return f"<class '{data.__class__.__name__}'>"
+        return f"{getattr(data, '__repr_type__', data.__class__.__name__)}"
```
iamogbz, line 268: Still not sure this should be added here as it's easy enough to override

src/syrupy/extensions/amber/serializer.py
noahnu: I believe this is to easily differentiate a `#` in a string/snapshot content vs. the snapshot separator. E.g.

```
assert snapshot == "#\n#\n#"
```

makes parsing easier, and I think it also makes it more readable

src/syrupy/filters.py
```diff
@@ -13,7 +13,7 @@ def paths(path_string: str, *path_strings: str) -> "PropertyFilter":
     Factory to create a filter using list of paths
     """
 
-    def path_filter(prop: "PropertyName", path: "PropertyPath") -> bool:
+    def path_filter(*, prop: "PropertyName", path: "PropertyPath") -> bool:
```
noahnu, line 16: (note) should capture this in a "migration" guide (which can also just be a longer comment in the changelog)

tests/syrupy/extensions/amber/__snapshots__/test_amber_filters.ambr
```diff
@@ -1,36 +1,36 @@
 # name: test_filters_error_prop[path_filter]
-  <class 'WithNested'> {
+WithNested(
```
noahnu, line 2: (note) Okay I can see how this is nicer 😄 

What happens if you have a multiple line string in the attribute though?

noahnu (commented):
I agree this looks nicer. I do think we should keep the initial indent though.

src/syrupy/extensions/amber/serializer.py
iamogbz: Good point will add a test to capture that

src/syrupy/filters.py
iamogbz: I could add it, but I didn't think there was a need as the function is already always called with keyword arguments. Also this function is returned by the factory, not really useful except when passed to the assertion

tests/syrupy/extensions/amber/__snapshots__/test_amber_filters.ambr
iamogbz: You can see an example in one of the test cases, where sets and tuples were used as keys [`test_amber_serializer::test_dict[actual2]`](https://github.com/tophat/syrupy/blob/fd878878d107d6fb5b57004e89452013f2f52935/tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr#L137). Will add a multiline string case too.

conftest.py
```diff
@@ -4,7 +4,7 @@
 
 from syrupy.utils import env_context
 
-typing.TYPE_CHECKING = True
+typing.TYPE_CHECKING = False
```
iamogbz, line 7: Mypy already validates type checking and `typing.Protocol` not available python <3.8. `Protocol` is used for defining types which is only used in development.

src/syrupy/extensions/amber/serializer.py
```diff
@@ -199,21 +199,16 @@ def serialize_string(cls, data: str, *, depth: int = 0, **kwargs: Any) -> str:
...
     ) -> str:
-        open_paren, close_paren = next(
-            parens
-            for iter_type, parens in {
-                GeneratorType: ("(", ")"),
-                list: ("[", "]"),
-                tuple: ("(", ")"),
-            }.items()
-            if isinstance(data, iter_type)
-        )
```
noahnu, line 210: What's the reason for dropping round brackets for the tuples?

noahnu (approved):

PR #508, opened 2021-06-01
chore: only run with codecov once

noahnu:


.github/workflows/cicd.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           python -m pip install -e . -r dev_requirements.txt
       - name: Test
         env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+          CODECOV_TOKEN: ${{ matrix.python-version == '3.9' ? secrets.CODECOV_TOKEN : '' }}
```
iamogbz, line 52: Suggestion just move the env and `--coverage` flag to the `version_checks` step instead

PR #519, opened 2021-06-11
chore: fix coverage reporting

noahnu:
Close #517

PR #521, opened 2021-06-12
Fix/codecov status

mcataford:
## Description

This is an attempt at fixing the Codecov status reporting.

Summary of changes:
- ~(Unrelated, will happily drop that one if need be) Use the [Codecov GHA](https://github.com/codecov/codecov-action) instead of the shell script approach. This feels less clunky and more sustainable.~ Trimming that out for now, doesn't seem to contribute to the issue at hand.
- Fixes the config layout. The `notify` key tripped the Codecov YAML validator (run using `cat codecov.yml | curl --data-binary @- https://codecov.io/validate`).

Now the validator :heavy_check_mark: the configuration file, which gives me hope. ~I have a sneaking suspicion that it's only looking at the config file if it's in `master` though.~ <-- Codecov docs mention that commit config can override the one in `master`, so we should be able to get reporting here.

## Related Issues

Follow-up to #519 

## Checklist

## Additional Comments

No additional comments.


mcataford:
Closing this, the issue was root in misconfiguration. No code change needed.

PR #523, opened 2021-06-13
test: get coverage reporting back to 💯 

iamogbz:
## Description

* Reverts noop changes to fix coverage
* Replaces `typing.Protocol` with `mypy_extensions.NamedArg` in typing
* Reenables typing in tests
* Enables coverage reporting on all build steps

## Checklist

## Additional Comments

No additional comments.


.github/workflows/cicd.yml
```diff
@@ -49,7 +48,9 @@ jobs:
           python -m pip install -U pip --no-cache-dir
           python -m pip install -e . -r dev_requirements.txt
       - name: Test
-        run: invoke test
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        run: invoke test --coverage
```
noahnu, line 53: Is this to capture the Windows case which might go down a different branch? Otherwise seems redundant to collect coverage multiple times.

.github/workflows/cicd.yml
```diff
@@ -58,8 +59,6 @@ jobs:
       SKIP_DEPS: 1
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
noahnu, line 62: I think you still need this for cov?

.github/workflows/cicd.yml
iamogbz: It's to cover both paths  where`mypy_extension` is and is not available e.g. in the minimum tests

.github/workflows/cicd.yml
iamogbz: I don't think so, at least from the behaviour in this repo. The `codecov` package seems to handle reporting correctly, issue was a misconfiguration on codecov website.

noahnu (approved):

PR #527, opened 2021-06-20
fix: support attrs v21 dependency

noahnu:


dev_requirements.txt
```diff
@@ -19,7 +19,9 @@ bleach==3.3.0
 certifi==2021.5.30
     # via requests
 cffi==1.14.5
-    # via pynacl
+    # via
+    #   cryptography
```
noahnu, line 23: This is only a dependency in ubuntu/linux, not Mac which is why we always bounce back and forth in this requirements file. It's a piptools limitation.

PR #531, opened 2021-08-02
fix: unused snapshot not filtered out when specific node

### LINKED ISSUE #529: Unused snapshot falsely detected when tests have same names
**Describe the bug**

If a test in one test file has the same name as a test in another test file, and both try to do a snapshot, if we run pytest but deselect one of those tests then pytest fails erroneously saying that a snapshot is obsolete.

**To reproduce**

Create two test files (`test_a.py` and `test_b.py`) with the following contents:

```py
def test_name(snapshot):
    assert snapshot == [1, 2, 3]


def test_other():
    assert True
```

Create the snapshots:

```
pytest --snapshot-update test_a.py test_b.py
```

Attempt to run all but `test_b.py::test_name`:

```
pytest test_a.py test_b.py::test_other
```

*Result:* pytest exits with status 1 saying:

```
--------------------------- snapshot report summary ---------------------------
1 snapshot passed. 1 snapshot unused.

```

**Expected behavior**

pytest should exit successfully and we shouldn't be told a snapshot is unused, because we didn't run the test that used it.

**Screenshots**

Should be clear from above.

- OS: several (macOS, Linux)
- Syrupy Version: 1.3.1
- Python Version: 3.6 and 3.7 (and I assume all the others)

**Additional context**

None

noahnu:
## Related Issues

- Closes #529

### DIFF
--- a/src/syrupy/report.py
+++ b/src/syrupy/report.py
@@ -198,8 +198,14 @@ class SnapshotReport:
                 unused_snapshots = {
                     snapshot
                     for snapshot in unused_snapshot_fossil
-                    if self._selected_items_match_name(snapshot.name)
-                    or self._provided_nodes_match_name(snapshot.name, provided_nodes)
+                    if self._selected_items_match_name(
+                        snapshot_location=snapshot_location, snapshot_name=snapshot.name
+                    )
+                    and self._provided_nodes_match_name(
+                        snapshot_location=snapshot_location,
+                        snapshot_name=snapshot.name,
+                        provided_nodes=provided_nodes,
+                    )
                 }
                 mark_for_removal = False
 
@@ -343,12 +349,22 @@ class SnapshotReport:
         ]
 
     def _provided_nodes_match_name(
-        self, snapshot_name: str, provided_nodes: List[List[str]]
+        self,
+        snapshot_location: str,
+        snapshot_name: str,
+        provided_nodes: List[List[str]],
     ) -> bool:
         """
-        Check that a snapshot name matches the node paths provided
+        Check that a snapshot name matches the node paths provided.
+        If no nodes are filtered, provided_nodes is empty, which means
+        all nodes should be matched.
         """
-        return any(snapshot_name in ".".join(node_path) for node_path in provided_nodes)
+        if not provided_nodes:
+            return True
+        for node_path in provided_nodes:
+            if snapshot_name in ".".join(node_path):
+                return True
+        return False
 
     def _provided_keywords_match_name(self, snapshot_name: str) -> bool:
         """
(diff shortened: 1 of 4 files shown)


src/syrupy/report.py
```diff
@@ -346,9 +346,14 @@ def _provided_nodes_match_name(
         self, snapshot_name: str, provided_nodes: List[List[str]]
     ) -> bool:
         """
-        Check that a snapshot name matches the node paths provided
+        Check that a snapshot name matches the node paths provided.
+        If no nodes are filtered, provided_nodes is empty, which means
+        all nodes should be matched.
```
noahnu, line 351: @iamogbz Is this statement correct?

noahnu: Just added another test which is failing, so this is definitely not true. We need to somehow filter provided_nodes so it's only considering relevant nodes. The problem is that the nodes are not fully qualified so they're ambiguous between files.

src/syrupy/report.py
```diff
@@ -198,8 +198,14 @@ def unused(self) -> "SnapshotFossils":
...
                     for snapshot in unused_snapshot_fossil
-                    if self._selected_items_match_name(snapshot.name)
-                    or self._provided_nodes_match_name(snapshot.name, provided_nodes)
+                    if self._selected_items_match_name(
+                        snapshot_location=snapshot_location, snapshot_name=snapshot.name
+                    )
+                    and self._provided_nodes_match_name(
+                        snapshot_location=snapshot_location,
+                        snapshot_name=snapshot.name,
+                        provided_nodes=provided_nodes,
```
noahnu, line 207: This bit might actually be dead code. It's never hit in tests. I don't think it's harmful at the moment, so will loop back to see if it can be removed, or if there's some edge case we're not testing.

src/syrupy/report.py
```diff
@@ -343,12 +349,22 @@ def _get_matching_path_nodes(self, snapshot_location: str) -> List[List[str]]:
...
+        If no nodes are filtered, provided_nodes is empty, which means
+        all nodes should be matched.
         """
-        return any(snapshot_name in ".".join(node_path) for node_path in provided_nodes)
+        if not provided_nodes:
+            return True
+        for node_path in provided_nodes:
+            if snapshot_name in ".".join(node_path):
+                return True
+        return False
```
noahnu, line 367: This syntax is easier to debug.

tests/integration/test_snapshot_similar_names_file_extension.py
```diff
@@ -0,0 +1,110 @@
...
+    result = testdir.runpytest(
+        "-v",
+        "--snapshot-details",
+        "--snapshot-default-extension",
+        "syrupy.extensions.single_file.SingleFileSnapshotExtension",
+        "test_1.py::test_b",
+        "test_2.py::test_a",
+    )
+    result.stdout.re_match_lines("2 snapshots passed")
+    assert result.ret == 0
```
noahnu, line 110: Duplicated this test file since locations for images are handled a bit differently, better safe than sorry.

Issue #539, opened 2021-08-16
Snapshot discovery and reading takes quadratic time


huonw:
**Describe the bug**

We're using syrupy, and it works well. Thank you!

Unfortunately we have nearly 500 snapshots, and our tests runs are starting to get quite slow. It seems like syrupy unfortunately makes the testing take quadratic time with respect to the number of snapshots.

**To reproduce**

Create this file
```python
# test_performance.py
import pytest
import os

SIZE = int(os.environ.get("SIZE", 1000))

@pytest.mark.parametrize("x", range(SIZE))
def test_performance(x, snapshot):
    assert x == snapshot
    # assert x == x
```

Run, for instance:
```
for s in 100 500 1000 2000; do
    echo "size = $s"
    # create the snapshots
    SIZE=$s pytest test_performance.py --snapshot-update
    # just check them
    SIZE=$s pytest test_performance.py
done
```

The times reported by pytest scale scales quadratically with the number of tests/snapshots (O(size**2)). I think this is because the number of `read_file`/`_read_snapshot_fossil` calls and `discover_snapshots` calls, as reported by `python -m cProfile -m py.test test_performance.py`, scales linearly (O(size)) with the number of tests/snapshots, and the work required for each call also scales linearly (because the files contain O(size) data, for the snapshots).

The times and number of calls (this is just for the invocation that's just checking snapshot) is something like:

| size | time (seconds) | `discover_snapshots` calls | `read_file` calls |
|---|---|---|---|
| 100 | 0.15 | 200 | 300 |
| 500 | 1.73 | 1000 | 1500 |
| 1000 | 6.24 | 2000 | 3000 |
| 2000 | 21.84 | 4000 | 6000 |

Things of note:

- each doubling from 500 -> 1000 -> 2000 multiplies the time by 4, approximately (classic marker of quadratic performance)
- the number of calls seems rather large: 2 discovery calls per test/snapshot, and 3 read-file calls

**Expected behavior**

The test runs should be linear in the number of tests/snapshots. For instance, if the `assert x == x` line is used (and the `snapshot` fixture removed) instead of `assert x == snapshot`, the test run is linear: even `SIZE=10000` finishes in < 4s on my machine.

It seems like this could be handled by discovering the snapshots once (or once per file) and reading each snapshot file once too.

**Screenshots**

**Environment (please complete the following information):**

- OS: macOS
- Syrupy Version: 1.4
- Python Version: 3.8

**Additional context**

[...]

PR #543, opened 2021-08-20
perf: memoise DataSerializer.read_file results

huonw:
## Description

This applies the `functools.lru_cache` memoising decorator to the low level `amber.DataSerializer.read_file` method, which dramatically improves performance by avoiding re-reading a single snapshot file dozens of times, thus eliminating the quadratic behaviour observed in #539 in the common case of read-only interaction with amber snapshots.

For the test in #539, with a single snapshot test parametrised with `SIZE` instances (note that doubling the size now doubles the time):

| SIZE | before (1.43) time (s) | after (this PR) time (s) |
|---|---|---|
| 100 | 0.09 | 0.07 | 
| 500 | 0.75 | 0.29 |
| 1000 | 2.42 | 0.51 |
| 2000 | 8.24 | 0.98 |
| 10000 | 201 | 4.64 |

For our real-world test suite, with 453 snapshots across 25 files:

| syrupy version | time (s) |
|---|---|
| 1.4.1 (original) | 68 |
| 1.4.2 | 51 |
| 1.4.3 | 35 |
| this PR | 30 |

This has to be on this `DataSerializer` class method, rather than the higher level methods in `AmberSnapshotExtension` because the methods there are instance methods, with a `self` argument, and it's called on different instances (and thus end up as different `lru_cache` entries).

This optimisation is limited to only reading the amber file types, but I think images (and other single-file snapshots) will not have the same quadratic issues, and writing is arguably much less common than reading.

## Related Issues

- #539

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/extensions/amber/serializer.py
+++ b/src/syrupy/extensions/amber/serializer.py
@@ -1,3 +1,4 @@
+import functools
 import os
 from types import GeneratorType
 from typing import (
@@ -12,7 +13,6 @@ from typing import (
     Tuple,
     Union,
 )
-import functools
 
 from syrupy.constants import SYMBOL_ELLIPSIS
 from syrupy.data import (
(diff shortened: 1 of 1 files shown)


noahnu (approved):

noahnu:
@all-contributors add @huonw for code, bug

PR #552, opened 2021-09-15
feat: add JSON snapshot extension

### LINKED ISSUE #551: Add extension for JSON snapshots

Usually I use snapshots when testing REST APIs. It's quite convenient to not compare response to a long dictionary, but rather check status code and that response is aligned with a snapshot. Quite often responses are JSON compatible. Therefore I would like to have in `syrupy` an extension supporting that.

```python
from datetime import datetime
 
import pytest

from syrupy.extensions.json import JSONSnapshotExtension
from syrupy.matchers import path_name_str

@pytest.fixture
def snapshot_json(snapshot):
    return snapshot.use_extension(JSONSnapshotExtension) 


def test_api_call(client, snapshot_json):
    resp = client.post("/endpoint")
    assert resp.status_code == 200
    assert snapshot_json == resp.json()


def test_create_user(client, snapshot_json):
    resp = client.post("/user", json={"name": "Jane"})
    assert resp.status_code == 201

    matcher = path_name_str({
      "id": (int,),
      "registeredAt": (datetime,)
    })
    assert snapshot_json(matcher=matcher) == resp.json()
```

Following tests should create `<test_name>.json` snapshots with following content:
```json
{
  "foo": "bar"
}
```

and in case of dynamic data:

```json
{
  "id": "<INT>",
  "registeredAt": "<DATETIME>",
  "name": "Jane"
}
```

If you have some concerns about having this bundled into `syrupy`, then I could create my own extension library, but to be honest I think the scenario I described is quite common, so I would appreciate it to be available in syrupy out of the box.

**Additional context**

Probably I should have created this FR before rushing to implementation, but there's a PR ready. I would consider it as a draft, and would appreciate to hear any feedback cause I might have missed something conceptually-wise. On the other hand, I mostly copy-pasted this extension from my current project, where it's already used and it works pretty good

**Motivation**

Theoretically, I can just use default Amber extension, but I don't like the format of it in this case. When your JSON response is a `.json` file, then you have nicer diffs, nicer syntax highlighting in your code editor, and most importantly, it's easier to just look into them

fbjorn:
## Description

This PR adds new `JSONSnapshotExtension` class, in order to make it possible to serialize JSON-like structures (dicts, lists) as `.json` files on disk.

It's also introducing new `path_name_str` matcher to make snapshots feel more "natural" when working with API responses. This is optional and just a proposal though.

## Related Issues

- Closes #551, Add JSONSnapshotExtension

## Checklist

- [x] This PR has sufficient test coverage. _(is 85% enough? I tried to cover most vulnerable parts)_
- [ ] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary). _I haven't bumped any versions yet, cause it's more like a draft_

## Additional Comments

Tested this extension in one of my projects already


N0124:
Does it work with arrays, that contains another dicts? And I want to exclude some key inside this inner dict?

fbjorn:
@N0124 funnily, I faced this problem a couple of weeks ago in my project. I had a response like 
```
{
  "items": [
    {
      "uuid": "abab-...",
      "createdDate": "...",
      "foo:" "bar",
      "foo2:" "bar2"
    }
  ],
  "page": 1
}
```
So I was trying to remove `uuid` and `createdDate` from objects in `items`. I tried to refactor the matcher from this PR to support some array syntax or something, but turned out it was easier just to mock `uuid` generation and `utc_now()` functions in my tests, so snapshots always contain real data, despite that it's mocked.

So answering to your question -- nope, currently there's no way to do it, but also I'm not sure it's possible with the current syntax `syrupy` provides for matchers

By the way sorry that this PR is hanging for a while, @noahnu does it make sense for you still to complete this PR if I remove `path_name_str` from the PR? So basically just adding a new snapshot extension for JSONs?

src/syrupy/extensions/json.py
```diff
@@ -0,0 +1,81 @@
...
+    @classmethod
+    def _filter(
+        cls,
+        data: SerializableData,
+        filtered: SerializableData,
+        path: PropertyPath,
+        exclude: Optional["PropertyFilter"] = None,
+        matcher: Optional[PropertyMatcher] = None,
+    ) -> SerializableData:
+        if not isinstance(data, dict):
```
N0124, line 38: I've refactored this function a little bit. I've tested it with exclude(props and paths) and matcher path_type - and it seems to work. What do you think?
```
        if not isinstance(data, dict):
            return data
        for k, v in data.items():
            if exclude and exclude(prop=k, path=path):
                continue
            path_tpl = (*path, (k, type(v)))
            if isinstance(v, dict):
                filtered[k] = {}
                filtered[k] = cls._filter(v, filtered[k], path_tpl, exclude, matcher)
                continue
            elif isinstance(v, list):
                filtered[k] = []
                for item in v:
                    result = cls._filter(item, {}, path_tpl, exclude, matcher)
                    filtered[k].append(result)
                continue
            if matcher:
                v = matcher(data=v, path=path_tpl)
            if isinstance(v, Repr):
                filtered[k] = repr(v)
            elif isinstance(v, datetime.datetime):
                filtered[k] = v.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
            else:
                filtered[k] = v
        return filtered
```

noahnu:
Feel free to pull out matcher changes into a separate PR

PR #561, opened 2021-10-24
feat: support custom snapshot names

### LINKED ISSUE #555: Give custom name to image snapshot file
Hello I have some tests where I need to generate multiple snapshot images in the same test.

Is it possible to provide the target snapshot name file to `snapshot` when asserting the images are equal ?

My use case:

```py
from syrupy.extensions.image import PNGImageSnapshotExtension
import base64
import pytest


@pytest.fixture
def snapshot(snapshot):
    return snapshot.use_extension(PNGImageSnapshotExtension)

def test_some_func(snasphot):
   base64_img1, base64_img2 = func()

    assert base64.base64decode(base64_img1) == snapshot
    assert base64.base64decode(base64_img2) == snapshot
```

It works as intended, but IMO the syntax is weird and magical, (we are basically testing 2 different things against the same object).

what I'm looking for / propose is to be able to set a custom filename for the snapshot if we want:

```py
    assert base64.base64decode(base64_img1) == snapshot(filename="photo_image.png")
    assert base64.base64decode(base64_img2) == snapshot(filename="signature_image.png")
```

Thanks!

bendidi:
## Description

This PR adds to possibility to specify a custom snapshot name suffix.

 Currently syrupy uses an `index` generated from `num_executions` of the snapshot fixture in a particular test instance, this PR adds to possibility to specify a string that will override that `index`.

## Related Issues

- Closes #555 .

## Checklist

## Additional Comments

This PR is not yet complete, there still work that needs to be done to add more tests and more examples with custom suffix for image snapshots. I want to validate the proposed solution before moving further with documentation and tests :)


bendidi:
Hello @noahnu any news on this ?
I may have a simple proposition if it interests you ^^

Why not update the `matches_snapshot_name` to only check that `self.__parse(snapshot_name).startswith(self.__parse(self.snapshot_name))`

noahnu:
Merged in #563

noahnu:
Available in syrupy 1.5.0

PR #563, opened 2021-10-26
feat: add support for custom snapshot names, close #555

### LINKED ISSUE #555: Give custom name to image snapshot file

noahnu:
## Description

Based on https://github.com/tophat/syrupy/pull/561.

## Related Issues

- Closes #555


noahnu:
For context: had to wrap the custom name in square brackets to prevent the "cannot relate snapshot name" warning. Reason this works is because https://github.com/tophat/syrupy/blob/818d405a85c2f1f5db9d673e632677c10cb52ad9/src/syrupy/location.py#L67 stops reading the name at the first non-valid python id character, which would be `[` in this case. Being able to relate snapshots is critical for discovering unused snapshots.

In the next major version of syrupy, I'd propose including the index in the snapshot itself, separate from the snapshot name so we can better differentiate the two.

noahnu:
@all-contributors add bendidi for code

PR #575, opened 2021-12-26
feat: release syrupy v2

noahnu:
# Syrupy v2

## Breaking Changes
- [x] Change amber serialization to be more python like (#505)
  - This changes the default amber serialization to better match python syntax. This does **NOT** need the file to be python parseable, just leverages python syntax recognition to make snapshots easier to grok.
- [x] Enforces keyword arguments in filter and matcher factory functions (#505)

## Deferred to v3
- [ ] Reclaim much of the API as internal implementation details so it's no longer part of the "public API contract"


zevisert:
I just spent a few minutes upgrading, but by adding `"*.ambr": "python"` to `"files.associations"`, vscode's languge server (pylance) really starts throwing a fit about all the syntax issues since these aren't meant to be valid python. @noahnu, do you have any suggestions on how to get language highlighting, without linting? A `# type: ignore` comment works, but that's an edit to the snapshot, and a path I think we generally don't want to go down. I tried `"python.linting.ignorePatterns"`, but it didn't seem to have any effect. 

The other idea I had was to set the language to something that parses amber _well enough_, that I don't have a language server for. Perl, Julia, Ruby, and Coffeescript all work well enough.

PR #587, opened 2022-01-25
fix(json): Use additional forward references for py3.7

dtczest:
* fixed 2 additional instances

## Description

Finishes fixing using syrupy with Python 3.7

## Related Issues

#586


noahnu:
@all-contributors add @dtczest for bug fix

PR #607, opened 2022-07-05
fix: ignore test_a_suffix snapshots when running test_a

### LINKED ISSUE #596: Unused snapshots when running on filename (test_a.py) that's the prefix of another (test_abc.py)
**Describe the bug**

I have a project that has filenames where the non-`.py` part happens to be a prefix of another (e.g. `test_a.py` vs. `test_abc.py`). When running `test_a.py` in isolation, syrupy picks up the snapshots from `test_abc.py` and either warns/errors about them or deletes them (depending on if `--snapshot-update`) is specified.

(Thanks for syrupy!)

**To reproduce**

```python
# test_a.py
def test_foo(snapshot): assert snapshot == "a"
```
```python
# test_abc.py
def test_bar(snapshot): assert snapshot == "abc"
```
```
# requirements.txt
attrs==21.4.0
colored==1.4.3
iniconfig==1.1.1
packaging==21.3
pluggy==1.0.0
py==1.11.0
pyparsing==3.0.8
pytest==7.1.1
syrupy==1.7.4
tomli==2.0.1
```

With the above code, if I run `pytest test_a.py`, the snapshot from `test_abc.py` is flagged as unused:

```
================================= test session starts =================================
platform darwin -- Python 3.9.10, pytest-7.1.1, pluggy-1.0.0
rootdir: .../syrupy-file-prefix
plugins: syrupy-1.7.4
collected 1 item                                                                      

test_a.py .                                                                     [100%]

------------------------------- snapshot report summary -------------------------------
1 snapshot passed. 1 snapshot unused.

Unused test_bar (__snapshots__/test_abc.ambr)
================================== 1 passed in 0.01s ==================================
```

Fully packaged:

```shell
# set up virtualenv/deps/files
python --version  # Python 3.9.10
python -m venv venv
. venv/bin/activate
pip install attrs==21.4.0 colored==1.4.3 iniconfig==1.1.1 packaging==21.3 pluggy==1.0.0 py==1.11.0 pyparsing==3.0.8 pytest==7.1.1 syrupy==1.7.4 tomli==2.0.1

echo 'def test_foo(snapshot): assert snapshot == "a"' > test_a.py
echo 'def test_bar(snapshot): assert snapshot == "abc"' > test_abc.py

# create snapshots
pytest --snapshot-update  # 2 snapshots generated.
pytest  # 2 snapshots passed.

# running test_abc.py: all okay
pytest test_abc.py --snapshot-details  # 1 snapshot passed.

# running test_a.py: unused snapshots (BUG)
pytest test_a.py --snapshot-details  # Unused test_bar (__snapshots__/test_abc.ambr)
pytest test_a.py --snapshot-details --snapshot-update  # Deleted test_bar (__snapshots__/test_abc.ambr)
```

**Expected behavior**

When running pytest against a single/subset of files, snapshots from other files shouldn't be considered unused.

**Screenshots**

- OS: macOS 12.3
- Syrupy Version: 1.7.4
- Python Version: 3.9.10

**Additional context**

N/A

huonw:
## Description

This fixes #596 by making the filename vs. snapshot location matching stricter: instead of allowing a filename of `test_file` to match if it's a substring of the location (e.g. `FOOtest_fileBAR` would match), it now only matches if either of these are true:

- the filename exactly matches, ignoring the extension (e.g. `test_file.ambr` matches, while `test_file_whatever.ambr` does not)
- the parent directory exactly matches (e.g. `.../test_file/foo.bar` matches, but `.../test_file.foo/bar` does not, nor `.../test_file/subdirectory/foo.bar`)

I've added tests for this, as well as expanding the existing ones for #529 to include checking test names that have matching suffices (especially relevant for `tests/integration/test_snapshot_similar_names_file_extension.py`, with the single file snapshots). All three sets of changed tests fail without the change to `location.py`.

## Related Issues

- Closes #596

## Checklist

## Additional Comments

Thanks for Syrupy!

### DIFF
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -114,4 +114,9 @@ class PyTestLocation:
         return self.__parse(self.snapshot_name) == self.__parse(snapshot_name)
 
     def matches_snapshot_location(self, snapshot_location: str) -> bool:
-        return self.filename in snapshot_location
+        loc = Path(snapshot_location)
+        # "test_file" should match_"test_file.ext" or "test_file/whatever.ext", but not
+        # "test_file_suffix.ext"
+        return self.filename == loc.stem or any(
+            self.filename == parent.name for parent in loc.parents
+        )
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -117,6 +117,4 @@ class PyTestLocation:
         loc = Path(snapshot_location)
         # "test_file" should match_"test_file.ext" or "test_file/whatever.ext", but not
         # "test_file_suffix.ext"
-        return self.filename == loc.stem or any(
-            self.filename == parent.name for parent in loc.parents
-        )
+        return self.filename == loc.stem or self.filename == loc.parent.name
--- a/tests/integration/test_snapshot_similar_names_default.py
+++ b/tests/integration/test_snapshot_similar_names_default.py
@@ -22,16 +22,18 @@ def testcases():
 @pytest.fixture
 def run_testcases(testdir, testcases):
     pyfile_content = "\n\n".join(testcases.values())
-    testdir.makepyfile(test_1=pyfile_content, test_2=pyfile_content)
+    testdir.makepyfile(
+        test_1=pyfile_content, test_2=pyfile_content, test_1_with_suffix=pyfile_content
+    )
     result = testdir.runpytest("-v", "--snapshot-update")
-    result.stdout.re_match_lines((r"4 snapshots generated\."))
+    result.stdout.re_match_lines((r"6 snapshots generated\."))
     return testdir, testcases
 
 
 def test_run_all(run_testcases):
     testdir, testcases = run_testcases
     result = testdir.runpytest("-v")
-    result.stdout.re_match_lines("4 snapshots passed")
+    result.stdout.re_match_lines("6 snapshots passed")
     assert result.ret == 0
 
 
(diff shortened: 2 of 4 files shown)


src/syrupy/location.py
```diff
@@ -114,4 +114,9 @@ def matches_snapshot_name(self, snapshot_name: str) -> bool:
         return self.__parse(self.snapshot_name) == self.__parse(snapshot_name)
 
     def matches_snapshot_location(self, snapshot_location: str) -> bool:
-        return self.filename in snapshot_location
+        loc = Path(snapshot_location)
+        # "test_file" should match_"test_file.ext" or "test_file/whatever.ext", but not
+        # "test_file_suffix.ext"
+        return self.filename == loc.stem or any(
+            self.filename == parent.name for parent in loc.parents
+        )
```
noahnu, line 122: I assume this is to handle the SingleFileExtension (image + json). Right now it's not strict enough since the parents will traverse all the way up to the root of the file system.

My proposed logic:

```suggestion
        return self.filename == loc.stem or self.filename == loc.parent.name
```

This does break the ability for an extension to have additional depth in their snapshot directories, but I think that's an okay compromise.

huonw: Sounds good to me 👍 I've made that update and validated it in the tests.

noahnu (approved):

huonw:
Woohoo, thank you!

PR #629, opened 2022-11-03
fix: lru_cache on snapshot reads

noahnu:


src/syrupy/extensions/amber/__init__.py
```diff
@@ -47,9 +47,10 @@ def _file_extension(self) -> str:
     def _read_snapshot_fossil(self, snapshot_location: str) -> "SnapshotFossil":
         return DataSerializer.read_file(snapshot_location)
 
+    @staticmethod
     @lru_cache()
     def __cacheable_read_snapshot(
-        self, snapshot_location: str, cache_key: str
```
noahnu, line 52: The "self" was being included in the lru_cache which effectively made the lru_cache pointless

PR #666, opened 2022-12-29
chore: update dev python version to 3.11

noahnu:
And update poetry + dependencies.


.github/workflows/cicd.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: ['3.8', '3.9', '3.10', '3.11-dev']
-      fail-fast: false
+      fail-fast: true
```
noahnu, line 35: Experimenting with this.

tasks/__init__.py
```diff
@@ -1,3 +1,8 @@
+import inspect
+
+if not hasattr(inspect, "getargspec"):
+    inspect.getargspec = inspect.getfullargspec
+
```
noahnu, line 5: https://github.com/pyinvoke/invoke/issues/833

Issue #535, opened 2021-08-03
Incompatible with pytest-xdist / thread safety


rassie:
**Describe the bug**

I'm using `pytest-xdist` to run my tests in parallel (mostly a single test parameterized for a vast amount of different test data). Since I wanted to try snapshot-based testing, I've added `syrupy` to the mix. After initial creation of `__snapshots__/mytest.ambr` (using the `--snapshot-update`), it disappears from disk at some point during the test run. Without `pytest-xdist` (i.e. without using `-n auto` option) this doesn't happen.

**To reproduce**

I don't have a clean reproduction, only an observation. My setup is a single test for a bit of processing, parameterized with a couple of hundreds different input files. While the `.ambr` file appear initially, after about 80% of tests done, the file disappears.  

**Expected behavior**

Snapshot files get updated with correct data without disappearing, even when used with `pytest-xdist` or `pytest-parallel`.

**Environment:**

- OS: Ubuntu 20.04
- Syrupy Version: 1.4.0
- Python Version: 3.8

**Additional context**

I assume the problem happens because each test execution writes to the file individually and thus a race condition happens. Not sure what can be done about it, maybe buffer the snapshot file somewhere and write it only at the end of the test run? There are surely better ideas. I assume that e.g. the JUnit reporter works correctly with `pytest-xdist`, maybe check what it does?

Either way, I wanted to bring this to your attention, would be happy if there was a short- to middle-term solution. If not, I'll run the suite without parallelization, it's not that big of a deal at the moment, at least for me :)

noahnu:
I'll need to dig into how pytest-xdist operates. Currently when in "write mode" (snapshot update flag passed) we write immediately when making the assertion. Buffering and deferring this to the session end hook where we already delete unused snapshots shouldn't be too difficult.

noahnu:
I'm re-examining where we do the reads/writes in an effort to solve https://github.com/tophat/syrupy/issues/539. I'm hoping that work will also solve the concurrency issue with pytest-xdist.

Sharu95:
@noahnu, I'm coming from https://github.com/pytest-dev/pytest/issues/9600, where I experienced the same issues as @rassie. Didn't quite manage to isolate the issue to whether it was syrupy, pytest or pytest-xdist at first, but good that I found this thread 👍 Do you have any updates of the progress on this issue/anywhere you left off that we might pick up to resolve this? :)

noahnu:
[...]

PR #667, opened 2022-12-29
fix: ensure all pytest options are serializable

noahnu:
Addresses https://github.com/tophat/syrupy/issues/535#issuecomment-950893723 by ensuring the default extension pytest option is serializable.

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -1,5 +1,6 @@
 import argparse
 import sys
+from functools import lru_cache
 from gettext import gettext
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


src/syrupy/__init__.py
```diff
@@ -30,13 +31,21 @@
...
+    if not value:
+        return DEFAULT_EXTENSION
     try:
         return import_module_member(value)
     except FailedToLoadModuleMember as e:
         raise argparse.ArgumentTypeError(e)
 
 
+def __default_extension_option(value: Optional[str]) -> Any:
+    __import_extension(value)
```
noahnu, line 45: We attempt to import the option to validate it as early as possible.

PR #676, opened 2023-01-15
refactor: simplify data serializer for ambr

noahnu:
BREAKING CHANGE: Serializers may now throw a TaintedSnapshotError which will tell the user to regenerate the snapshot even if the underlying data has not changed. This is to support rolling out more subtle changes to the serializers, such as the introduction of serializer metadata.

BREAKING CHANGE: Renamed DataSerializer to AmberDataSerializer.


src/syrupy/extensions/amber/serializer.py
```diff
@@ -87,43 +90,60 @@ def write_file(
...
-                            snapshot_collection.add(
-                                Snapshot(
+                    if line.startswith(cls.Marker.Comment):
+                        marker_key, *marker_rest = line[marker_offset:].split(
+                            ":", maxsplit=1
+                        )
+                        marker_key = marker_key.rstrip(" \r\n")
+                        marker_value = marker_rest[0] if marker_rest else None
+
+                        if marker_key == cls.Marker.Name:
```
noahnu, line 118: Makes it easier to reason about arbitrary markers. Planning on introducing a new marker for index shortly

tests/syrupy/extensions/amber/__snapshots__/test_amber_matchers.ambr
```diff
@@ -1,3 +1,4 @@
+# serializer version: 1
```
noahnu, line 1: In theory, this may let us support multiple concurrent versions of a serializer without requiring a major release.

Issue #657, opened 2022-11-30
Tests with many snapshots are sorted 1, 10, 11, ..., 2, 20, ..., 3, 4, 5, ..., not numeric order


huonw:
**Describe the bug**

Snapshots are stored sorted in the `.ambr`, but this seems to be doing a string comparison on the index, meaning `10` sorts before `2`, etc.

**To reproduce**

```python
# test_many.py
def test_many(snapshot):
    for i in range(25):
        assert i == snapshot
```

Run `pytest --snapshot-update test_many.py`:

Resulting `__snapshots__/test_many.ambr`:

```
# name: test_many
  0
# ---
# name: test_many.1
  1
# ---
# name: test_many.10
  10
# ---
# name: test_many.11
  11
# ---
# name: test_many.12
  12
# ---
# name: test_many.13
  13
# ---
# name: test_many.14
  14
# ---
# name: test_many.15
  15
# ---
# name: test_many.16
  16
# ---
# name: test_many.17
  17
# ---
# name: test_many.18
  18
# ---
# name: test_many.19
  19
# ---
# name: test_many.2
  2
# ---
# name: test_many.20
  20
# ---
# name: test_many.21
  21
# ---
# name: test_many.22
  22
# ---
# name: test_many.23
  23
# ---
# name: test_many.24
  24
# ---
# name: test_many.3
  3
# ---
# name: test_many.4
  4
# ---
# name: test_many.5
  5
# ---
# name: test_many.6
  6
# ---
# name: test_many.7
  7
# ---
# name: test_many.8
  8
# ---
# name: test_many.9
  9
# ---
```

**Expected behavior**

Snapshots should be sorted in numeric order 1, 2, ..., 9, 10, ..., 20, ..., 24. 

Some options:

- formatting the number with leading zeros. If it's possible the leading zeros could be computed dynamically based on how many snapshots are actually used in each test, otherwise could be hard-coded (e.g. 01, 02, ... 09, 10, ... would ensure natural sorting up to 100 snapshots in one test)
- not doing string comparison, e.g. https://github.com/tophat/syrupy/issues/569 might be related.

**Screenshots**

N/A

**Environment (please complete the following information):**

- OS: macOS
- Syrupy Version: 3.0.5
- Python Version: 3.9

**Additional context**

Thank you for syrupy!

noahnu:
I like the idea of implementing https://github.com/tophat/syrupy/issues/569 first, removing the index from the name field and then sorting by (name, index).

[...]

PR #680, opened 2023-01-26
fix: improve performance of amber sort

noahnu:
https://github.com/tophat/syrupy/issues/657#issuecomment-1404456390


noahnu:
Not seeing a perf improvement.

PR #681, opened 2023-01-26
fix: improve performance of snapshot sorting

noahnu:


src/syrupy/extensions/amber/serializer.py
```diff
@@ -88,24 +88,14 @@ class Marker:
...
-            return (0, part)
-
     @classmethod
     def __snapshot_sort_key(
         cls, snapshot: "Snapshot"
     ) -> List[Tuple[int, Union[str, int]]]:
-        return [cls.__maybe_int(part) for part in snapshot.name.split(".")]
+        if "." in snapshot.name:
+            index = snapshot.name.rfind(".") + 1
+            return f"{snapshot.name[:index-1]}.{snapshot.name[index:].zfill(4)}"
```
noahnu, line 97: The zfill is arbitrary but should ensure sorting of at least 1000 assertions within a single test case.

noahnu:
This PR doesn't solve the perf issues. Will have to do some actual profiling.

PR #733, opened 2023-04-06
Fix bootstrap script run in non-Bash shells

john-kurkowski:
## Description

The script uses some Bash-specific syntax. Previously, if the user's shell defaults to e.g. Zsh, the script would run and report syntax errors.

## Checklist

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Based on conventions from https://github.com/github/scripts-to-rule-them-all
 # script/bootstrap: Resolve all dependencies that the application requires to
 # run.


noahnu:
Just confirming, you're sourcing the script? It's meant to be sourced, not executed. I don't believe the shebang does anything when sourced.

john-kurkowski:
Oh shoot, you're right. The README says `. script/bootstrap`, not `script/bootstrap`. The former runs fine without the shebang (after I enabled Bash autocomplete compatibility in Zsh). I ran the latter to keep dependencies up to date, thinking that's how the script was supposed to be used, forgetting the README instructions.

john-kurkowski:
So, besides the minor confusion (do any scripts make it clear whether they're supposed to be sourced vs. executed?), I'd say things are working as intended. Never mind!

Issue #749, opened 2023-05-29
Crash with SingleFileSnapshotExtension


rotu:
**Describe the bug**

`SingleFileSnapshotExtension` crashes when trying to serialize a Dict with a number value.

**To reproduce**

run `pytest --snapshot-update` on a file containing the below test:

```python
from syrupy.extensions.single_file import SingleFileSnapshotExtension
from syrupy.extensions.amber import AmberSnapshotExtension
def test_foo(snapshot):
    value = {"x": 0}
    # this works:
    assert value == snapshot(extension_class=AmberSnapshotExtension)
    #this does not:
    assert value == snapshot(extension_class=SingleFileSnapshotExtension)
```

**Expected behavior**

The test should generate a snapshot file without errors.

**Screenshots**

```
>       assert value == snapshot(extension_class=SingleFileSnapshotExtension)
E       assert [+ received] == [- snapshot]
E         TypeError: 'str' object cannot be interpreted as an integer
E         Traceback (most recent call last):
E           File ".../site-packages/syrupy/assertion.py", line 268, in _assert
E             serialized_data = self._serialize(data)
E                               ^^^^^^^^^^^^^^^^^^^^^
E           File ".../site-packages/syrupy/assertion.py", line 181, in _serialize
E             return self.extension.serialize(
E                    ^^^^^^^^^^^^^^^^^^^^^^^^^
E           File ".../site-packages/syrupy/extensions/single_file.py", line 52, in serialize
E             return self.get_supported_dataclass()(data)
E                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

test_wireshark.py:24: AssertionError
```

**Environment (please complete the following information):**

- OS: macOS 13.5
- Syrupy Version: 4.0.2
- Python Version: CPython 3.11.3

noahnu:
The single file snapshot extension is meant for string or byte data (like images). If you want to store a dict, you'll have to encode it somehow. Alternatively you can use the JSON extension which extends the single file extension.

rotu:
> The single file snapshot extension is meant for string or byte data

Gotcha. That makes sense, though it wasn't obvious to me. Here are a few places that I think this should be fixed:

The documentation makes it sound like the difference is just whether it creates one or multiple files:

> - **`AmberSnapshotExtension`**: This is the default extension which generates `.ambr` files. Serialization of most data types are supported.
...
[...]

PR #754, opened 2023-06-03
Decouple different snapshot functionality

rotu:
Decouple different snapshot functionality

## Description

e.g. The plugin architecture is defined with multiple subparts. This plugin allows implementing those separately - e.g. you can use a different serializer from comparator.

## Related Issues

- Addresses use-case mentioned in #749, using default serializer but separate per-test snapshot files.

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -23,7 +23,13 @@ from .exceptions import (
 from .extensions.amber.serializer import Repr
 
 if TYPE_CHECKING:
-    from .extensions.base import AbstractSyrupyExtension
+    from .extensions.base import (
+        AbstractSyrupyExtension,
+        SnapshotCollectionStorage,
+        SnapshotComparator,
+        SnapshotReporter,
+        SnapshotSerializer,
+    )
     from .location import PyTestLocation
     from .session import SnapshotSession
     from .types import (
@@ -91,6 +97,19 @@ class SnapshotAssertion:
         init=False,
         default_factory=list,
     )
+    _serializer: Optional["SnapshotSerializer"] = field(init=False, default=None)
+    _storage: Optional["SnapshotCollectionStorage"] = field(
+        init=False,
+        default=None,
+    )
+    _comparator: Optional["SnapshotComparator"] = field(
+        init=False,
+        default=None,
+    )
+    _reporter: Optional["SnapshotReporter"] = field(
+        init=False,
+        default=None,
+    )
 
     def __post_init__(self) -> None:
         self.session.register_request(self)
@@ -105,6 +124,22 @@ class SnapshotAssertion:
         if not self._extension:
             self._extension = self.__init_extension(self.extension_class)
         return self._extension
+
+    @property
+    def serializer(self) -> "SnapshotSerializer":
+        return self._serializer or self.extension
+
+    @property
+    def storage(self) -> "SnapshotCollectionStorage":
+        return self._storage or self.extension
+
+    @property
+    def comparator(self) -> "SnapshotComparator":
+        return self._comparator or self.extension
+
+    @property
+    def reporter(self) -> "SnapshotReporter":
+        return self._reporter or self.extension
 
     @property
     def num_executions(self) -> int:
@@ -178,7 +213,7 @@ class SnapshotAssertion:
         assert self == data
 
     def _serialize(self, data: "SerializableData") -> "SerializedData":
-        return self.extension.serialize(
+        return self.serializer.serialize(
             data, exclude=self._exclude, matcher=self.__matcher
         )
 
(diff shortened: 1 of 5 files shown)


rotu:
This might be a very bad idea, but it also seems like it might be in line with the original plugin architecture intent. Curious your thoughts.

rotu:
> Will give this more thought at some point this week. Initial thought is that this should be doable with a custom extension like so (not tested, but I think this should work)
> 
> ```python
> from syrupy.extensions.amber import AmberDataSerializer
> from syrupy.extensions.single_file import SingleFileSnapshotExtension
> 
> class SingleFileAmberSnapshotExtension(SingleFileSnapshotExtension):
>     def serialize(self, data, **kwargs) -> "SerializedData":
>         return AmberDataSerializer.serialize(data, **kwargs)
> ```

I think you're right but it's very confusing to me how the library is set up. It seems like there are two broad philosophies to plugins:

1. Inherit from the base thingy and override certain functions with a new implementation.
2. Inject new implementation into well-defined places for hooks.

Inheriting from an extension and overriding seems weird to me because it's doing both.

-----

That said, there's a lot I don't understand about this codebase which prevents me from being confident in this PR:
- what "amber" means in "AmberSnapshot" and why it's used instead of pickle (I'm guessing because pickle is not text-based).
- why `snapshot` takes `extension_class` instead of `extension`
- why some extension methods like `get_snapshot_name`, `write_snapshot` are class methods instead of instance methods
- why `SnapshotAssertion` mutates itself in `_post_assert`.

noahnu:
> Inheriting from an extension and overriding seems weird to me because it's doing both.

I agree with that. There's definitely room to improve here. The codebase has already gone through several refactors, though it's pretty stable right now so I'd be hesitant to change anything too drastic without good reason. We also want to reduce the number of breaking changes if we can help it.

> what "amber" means in "AmberSnapshot" and why it's used instead of pickle (I'm guessing because pickle is not text-based).

Amber is a custom format we built intended to reduce the size of diffs and be easy to review in pull requests. We don't need the ability to deserialize, only serialize. The name is a play on words with the theme of "syrupy" that used to be more prevalent in the codebase. There was a strong Jurassic Park theme (syrupy = amber / fossils etc). We've expunged most of this from the codebase to improve readability. 

> why snapshot takes extension_class instead of extension

The extension defines an interface so to speak. It does not have any instance state.

> why some extension methods like get_snapshot_name, write_snapshot are class methods instead of instance methods

There was some recent refactoring to support pytest-xdist. Moving off of instances made it a lot simpler to reason about. Helps with performance as well. We cache some function calls using the extension class type. Not something we could easily do if there was instance data we had to consider.

> why SnapshotAssertion mutates itself in _post_assert.

It lets you customize specific assertions. The post_assert resets the snapshot fixture in between assertions. Lets you do things like:

```
def test_case(snapshot):
    assert x == snapshot(exclude=props("uuid"))
    assert y == snapshot(exclude=props("other_prop"))
```

noahnu:
Just posting to let you know I haven't forgotten about this. I do agree there's plenty of opportunity to simplify the architecture. I'd prefer to try more incremental changes in advance of any larger public API change (to reduce breaking changes and adoption pains).

I'm not sure I have too much time over the next few months to support _large scale_ changes though.

FWIW, a lot of the initial implementation was based on attrs before dataclasses, and when we moved to dataclasses we only did the bare minimum so it was functional. Leaning on newer patterns to simplify the code and make it easier to contribute to is definitely something I'm interested in.

And if you're curious, you were 100% right about the initial design for the extensions/plugins being closer to the `snapshot(storage=..., serializer=...)` model (if you look at some of the earlier commits in this repo, you'll see something a bit closer to that).

StephanMeijer:
> Will give this more thought at some point this week. Initial thought is that this should be doable with a custom extension like so (not tested, but I think this should work)
> 
> ```python
> from syrupy.extensions.amber import AmberDataSerializer
> from syrupy.extensions.single_file import SingleFileSnapshotExtension
> 
> class SingleFileAmberSnapshotExtension(SingleFileSnapshotExtension):
>     def serialize(self, data, **kwargs) -> "SerializedData":
>         return AmberDataSerializer.serialize(data, **kwargs)
> ```

Only `.encode()` at the end sees to be missing. Only issue: the Amber Extension gives proper diffs. How to get those back?

StephanMeijer:
The following snippet works beautifully:

```py
class SingleFileAmberSnapshotExtension(SingleFileSnapshotExtension):
    _file_extension = 'ambr'
    _write_mode = WriteMode.TEXT

    def serialize(self, data, **kwargs):
        return AmberDataSerializer.serialize(data, **kwargs)
  ```

noahnu:
Closing this PR as stale. The larger architectural discussions I'm still open to discuss, however I think we'd need a strong argument to make core architectural changes at this point (the codebase is fairly stable at the moment).

PR #710, opened 2023-02-16
fix: added support for custom file extension with multiple parts

tolgaeren:
## Description

This PR modifies the `single_file.py` to add support for multiple extension parts. An example is setting `"tar.gz"` as the custom `_file_extention` string:

```python
class DotInFileExtension(SingleFileSnapshotExtension):
    _file_extension = "tar.gz"
```

Previously this was causing some unexpected behavior as described in #703 . I've found the problematic part in `SingleFileSnapshotExtension._read_snapshot_collection` in `single_file.py`:

```python
 snapshot_collection.add(Snapshot(name=Path(snapshot_location).stem))
```

`.stem` only removes the last extension not all the extensions as [documented](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem). We would like to remove as many extensions as there are in the `_file_extension` string.

There are a few ways to fix it:
1. Calling `.stem` as many times as number of extensions.
2. Calling `.with_suffix("")` as many times as the number of extensions.
3. Replacing the `self._file_extension` at the end of  `snapsot_location` with `""`.

I tried option 3 first, it seemed to break a lot of other tests. Then I picked option 1 as it seemed it was the closest to original implementation. Let me know if you think of a better way to do it.

Added a test under `test_single_file.py` as it seemed like the right location for it.

## Related Issues

- Closes #703 , a short description of the linked issue.

## Checklist

- [x] This PR has sufficient documentation.
- [x] This PR has sufficient test coverage.
- [x] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary).

## Additional Comments

No additional comments.

PR #761, opened 2023-06-19
fix: incorrect marking of TestClass.test_method as unused, close #717

### LINKED ISSUE #717: Syrupy is recognizing snapshot as not being used (4.0.1)
**Describe the bug**

This is probably a regression in the latest release. I didn't observe the behavior in 4.0.0:

1. On a new test with no snapshot (using the `json` extension), I get the expected behaviour:
> 1 snapshot failed
> Snapshot 'TestFoo.test_foo' does not exist!
2. With the first `pytest --snapshot-update` (this is where the first signs of the bug is):
> 1 snapshot generated. 1 unused snapshot deleted.
Deleted TestFoo (tests/foo/__snapshots__/test_foo/TestFoo.test_foo.json)

Why does it output `Deleted ....`, that isn't the expected behaviour (the deletion doesn't happen though)
3. Another follow up when running `pytest`:
> 1 snapshot passed. 1 snapshot unused.

Again, that isn't correct. It gets worse though:
4. `pytest --snapshot-update`:
This time the deletion *will* happen hence the next run of `pytest` will fail since it won't find a snapshot.

# Env: 

Syrupy 4.0.1
Python 3.10

Going back to 4.0.0 solved the issue.

### LINKED ISSUE #752: Snapshots using JSONSnapshotExtension are deleted when they should not be
**Describe the bug**

I migrated from `AmberSnapshotExtension` to `JSONSnapshotExtension`. Updating after the initial conversion I get
```
9 snapshots generated. 9 unused snapshots deleted.
```
There are 9 snapshots, it regenerates all of them and then deletes them. 

In another case it would delete all 9, but then recreate 1 but in not in the `__snapshots__` folder directly, but in a folder `__snapshots__/test_filename` where `test_filename` is the name of the test script. The snapshot itself had the correct filename.

**To reproduce**

1. Create tests with `JSONSnapshotExtension`
2. Run `pytest --snapshot-update` and commit
3. Rerun  `pytest --snapshot-update` without actually having made any changes.
4. Get `9 snapshots generated. 9 unused snapshots deleted.` and indeed all your snapshots deleted.

**Expected behavior**

Not deleting snapshots that are still being used.

**Screenshots**

- OS: NixOS 22.11
- Syrupy Version: 4.0.2
- Python Version: 3.11.3

**Additional context**

noahnu:
## Description

Fixes regression from https://github.com/tophat/syrupy/pull/710

Closes #717, closes #752

### DIFF
--- a/src/syrupy/extensions/single_file.py
+++ b/src/syrupy/extensions/single_file.py
@@ -82,9 +82,10 @@ class SingleFileSnapshotExtension(AbstractSyrupyExtension):
     ) -> "SnapshotCollection":
         file_ext_len = len(self._file_extension) + 1 if self._file_extension else 0
         filename_wo_ext = snapshot_location[:-file_ext_len]
+        basename = Path(filename_wo_ext).parts[-1]
 
         snapshot_collection = SnapshotCollection(location=snapshot_location)
-        snapshot_collection.add(Snapshot(name=Path(filename_wo_ext).stem))
+        snapshot_collection.add(Snapshot(name=basename))
         return snapshot_collection
 
     def _read_snapshot_data_from_location(
--- a/tests/integration/test_single_file_multiple_extensions.py
+++ b/tests/integration/test_single_file_multiple_extensions.py
@@ -38,3 +38,39 @@ def test_multiple_file_extensions(testdir):
     result.stdout.re_match_lines((r"1 snapshot passed\."))
     assert "snapshots unused" not in result.stdout.str()
     assert result.ret == 0
+
+
+def test_class_style(testdir):
+    """
+    Regression test for https://github.com/tophat/syrupy/issues/717
+    """
+
+    testcase = """
+    import pytest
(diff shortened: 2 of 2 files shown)


src/syrupy/extensions/single_file.py
```diff
@@ -82,9 +82,10 @@ def _read_snapshot_collection(
     ) -> "SnapshotCollection":
         file_ext_len = len(self._file_extension) + 1 if self._file_extension else 0
         filename_wo_ext = snapshot_location[:-file_ext_len]
+        basename = Path(filename_wo_ext).parts[-1]
 
         snapshot_collection = SnapshotCollection(location=snapshot_location)
-        snapshot_collection.add(Snapshot(name=Path(filename_wo_ext).stem))
```
noahnu, line 87: We're already stripping the suffix on L84. We should not be doing this twice, otherwise our class method naming convention `TestClass.test_method.json` will turn into `TestClass`.

Issue #786, opened 2023-08-16
How to use Syrupy to save custom objects without converting them to strings


atharva-2001:
Dear Syrupy Team,
Thank you for this great package. I wanted to extend the `SingleFileSnapshotExtension` to save Pandas and NumPy objects to HDF files and `.dat` files respectively. I used the functions that you mentioned could be overridden(except the `serialize` function) to save these files in these new formats. However [this specific line](https://github.com/tophat/syrupy/blob/f4bc8453466af2cfa75cdda1d50d67bc8c4396c3/src/syrupy/assertion.py#L209) in `get_assert_diff` is causing an error that I just can't get past.

Both Dataframes and Arrays have no bool values, which is why doing something like `dataframe or ""` raises a [ValueError](https://stackoverflow.com/questions/10062954/valueerror-the-truth-value-of-an-array-with-more-than-one-element-is-ambiguous).

Any help on how I can get past this error? Am I doing the wrong thing?
Here are a few references-
![image](https://github.com/tophat/syrupy/assets/55894364/3b4dc7d6-e56a-4b50-8f05-34bed429ea6b)

Here is the extension that I wrote-
https://github.com/atharva-2001/syrupy-pandas-numpy/blob/main/tests/test_pd.py

noahnu:
That line can be updated to do an explicit "None" check. Are you interested in contributing / opening a PR?

atharva-2001:
Yes I would be very happy to! I would open a PR in sometime today!

atharva-2001:
@noahnu Would you mind looking at https://github.com/tophat/syrupy/pull/789? Is the change alright? Also could you please review my code a bit as well? Is it safe to extend the class in the way I did from a sustainability perspective? 

We are planning to add this library for our tests in TARDIS- https://tardis-sn.github.io/, https://github.com/tardis-sn and might be doing this across several repos.

Could the classes I made stay as is for at least 3-4 years?

noahnu:
Yes the API you're using is stable. I was thinking about renaming some of the methods to remove the underscore, however I have no immediate plans to do so, and if that happens, it'd be in a new major version with a migration guide.

PR #789, opened 2023-08-17
fix: preserve Falsy values in assertion diff function

atharva-2001:
## Description

This PR does a None check here-https://github.com/tophat/syrupy/blob/f4bc8453466af2cfa75cdda1d50d67bc8c4396c3/src/syrupy/assertion.py#L209

This helps me extend the `SingleFileSnapshotExtension` to save snapshots as HDF files. Please see issue https://github.com/tophat/syrupy/issues/786 to  know more.

## Related Issues

Closes https://github.com/tophat/syrupy/issues/786

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -206,7 +206,9 @@ class SnapshotAssertion:
             # Rotate to place exception with message at first line
             return lines[-1:] + lines[:-1]
         snapshot_data = assertion_result.recalled_data
-        serialized_data = assertion_result.asserted_data or ""
+        if assertion_result.asserted_data is None:
+            assertion_result.asserted_data = ""
+        serialized_data = assertion_result.asserted_data
         diff: List[str] = []
         if snapshot_data is None:
             diff.append(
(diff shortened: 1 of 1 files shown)


src/syrupy/assertion.py
```diff
@@ -206,7 +206,9 @@ def get_assert_diff(self) -> List[str]:
             # Rotate to place exception with message at first line
             return lines[-1:] + lines[:-1]
         snapshot_data = assertion_result.recalled_data
-        serialized_data = assertion_result.asserted_data or ""
+        if assertion_result.asserted_data is None:
+            assertion_result.asserted_data = ""
```
noahnu, line 210: We should avoid mutating the assertion result and instead should conditionally set serialized_data, like so:

```
serialized_data = assertion_result.asserted_data if assertion_result.asserted_data is not None else ""
```

noahnu:
@all-contributors add @atharva-2001 for code

atharva-2001:
Thank you!!

Issue #784, opened 2023-08-14
"exclude" filters don't appear to work correctly with msgspec structs


BenGatewood:
**Describe the bug**

When I try and exclude attributes from a custom class that inherits from [`msgspec.Struct`](https://jcristharif.com/msgspec/structs.html) the syntax/approach from the [docs](https://github.com/tophat/syrupy#exclude) doesn't seem to produce the desired behaviour and the excluded attributes remain in the snapshot

**To reproduce**

```import msgspec

from syrupy.filters import props


class Base(msgspec.Struct):
    foo: str
    bar: int


def test_base(snapshot):
    b = Base(
        foo="baz",
        bar=5,
    )

    assert b == snapshot(exclude=props("bar"))
```

**Expected behavior**

From the above example, I would expect a snapshot like: 
```# serializer version: 1
# name: test_base
  Base(foo='baz')
# ---
```

**Environment (please complete the following information):**

- OS: Mac OSX 13.5
- Syrupy Version: 4.0.8
- Python Version: 3.11

**Additional context**

Thanks for looking!

noahnu:
Custom classes are serialized by calling `repr`, see: https://github.com/tophat/syrupy/blob/2ed84ea8e2b9dfa01cd4788f9787809ffc6b2648/src/syrupy/extensions/amber/serializer.py#L373

Since repr returns a string, we don't have a way to modify the result of calling repr to apply an exclude. You could extend the serializer to support msgspec.Struct:

```py
import pytest
from typing import Any
from syrupy.filters import props
from syrupy.extensions.amber import AmberSnapshotExtension, AmberDataSerializer

import msgspec

class MySpec(msgspec.Struct):
    a: str
    b: int


class ExtendedAmberDataSerializer(AmberDataSerializer):
    @classmethod
    def serialize_unknown(cls, data: Any, *, depth: int = 0, **kwargs: Any) -> str:
        if isinstance(data, (msgspec.Struct,)):
            return data.__class__.__name__ + super().serialize_dict(
                data=msgspec.to_builtins(data),
                **kwargs
            )[len("dict"):]
        return super().serialize_unknown(data, depth=depth, **kwargs)


class ExtendedAmberSnapshotExtension(AmberSnapshotExtension):
    serializer_class = ExtendedAmberDataSerializer


@pytest.fixture
def snapshot(snapshot):
    return snapshot.use_extension(ExtendedAmberSnapshotExtension)

def test_case(snapshot):
[...]

Issue #547, opened 2021-08-28
[Feature Request] Regex matcher


j7zAhU:
The breaking factor seems to be the last test in my file, where I serialize a dataframe :

```
string = df.to_string()
len(string)   # 2_000
assert string == snapshot
```
When I run `pytest` at the root of the test dir, I get an assertion error and the terminal displays the chunk of text.

When I selectively leave out tests, I can get this one to work. So it's not that particular serialization.

noahnu:
Can you provide any more information, perhaps a repo where I can reproduce the issue? Or even a complete snippet of code I can copy and paste into a project to run?

Considering we have tests which cover multiple assertions and selectively running test files, from your description it sounds like it should work.

j7zAhU:
Sure. I'll try to put something together

noahnu:
> I get an assertion error and the terminal displays the chunk of text

It's telling you what changed. This is likely a real test failure, not a bug of any sort. The reason you're getting a whole chunk of text, is likely due to the fact that you're converting your Dict to a string before snapshotting it. That's not necessary, and means you lose out on ambr formatting, and more useful diffs.

So instead of:

```py
def test_case(snapshot):
    output_dir_structure = { key: "value" }
    assert json.dumps(outdir_dir_structure) == snapshot
```

do:

```py
def test_case(snapshot):
    output_dir_structure = { key: "value" }
    assert outdir_dir_structure == snapshot
```

If the issue is not obvious after switching away from a string, you'll need to step through the 2 tests which conflict. You likely have some sort of test pollution, where one test is affecting the other, whether it's the use of an auto-incrementer, some shared state, or improper "cleanup".

I also noticed you're snapshotting some file system paths. That's "non-deterministic" and means your snapshot will fail if run on any other machine, such as for another developer or on CI. In this case, you're interested in syrupy matchers: https://github.com/tophat/syrupy#matcher

j7zAhU:
Thank you. 

1. In the test file, I am serializing a Pandas dataframe. If I do not use the .to_string() method, Syrupy saves each element separately which is hard to read.

2. Regarding the matcher: I did run into the problem of non-deterministic file-paths, because pytest gives you a unique numbered directory like this: `tmp_path = C:\Users\X\AppData\Local\Temp\pytest-of-X\pytest-684\test_sample0` 

If I am building a test path such as `my_path = tmp_path / "my_subdir"`, how exactly can I ignore everything to the left of the 684?

Regards

[...]

PR #791, opened 2023-08-19
feat(serializer): add snapshot regex value matcher and bypass custom repr helper

iamogbz:
## Description

* Adds function `bypass_custom_repr` to help with serializing objects with custom third party class `repr` implementations
* Adds the `replacer` parameter to the `path_type` matcher signature to allow custom replacement actions on matched property values
* Adds matcher helper `path_value` to support regex match replacement on data values
* Deprecates the `regex` flag from the matcher to simplify the api, since the matcher and string compare serve equivalent functions

## Related Issues

- #547, adds regex matcher helper
- #784, helps make use-case simpler

## Checklist

## Additional Comments

No additional comments


poetry.lock
noahnu, line 1: Possibly a mismatch in poetry versions?

src/syrupy/extensions/amber/serializer.py
```diff
@@ -348,6 +349,17 @@ def serialize_unknown(cls, data: Any, *, depth: int = 0, **kwargs: Any) -> str:
...
+    @classmethod
+    def object_attrs(cls, data: Any) -> "Iterable[str]":
+        return (name for name in dir(data) if not name.startswith("_"))
+
+    @classmethod
+    def bypass_custom_repr(cls, data: Any) -> "tuple[Any, ...]":
+        attr_names = list(cls.object_attrs(data))
+        return collections.namedtuple(data.__class__.__name__, attr_names)(
+            **{prop: getattr(data, prop) for prop in attr_names}
+        )
```
noahnu, line 361: I was thinking about how the namedtuple serialization is exactly what we want when it came to https://github.com/tophat/syrupy/issues/784.

Regarding the name of the method, this does not really "bypass" custom repr, it's just a utility for constructing a named tuple based on an object's attrs. If it's "bypassing" some serialization, I'd expect it to somehow "communicate" with the serializer itself. I.e. a model where it sets a marker on an object so that the serializer knows to handle it differently.

Instead, I'd rename this to more explicitly be a utility that constructs a named tuple from a custom object. Perhaps `def make_named_tuple` or `def create_named_tuple` / `def convert_obj_to_named_tuple`.

If the idea is to use this in conjunction with the replacer to provide a more elegant solution to the msgspec serializer issue, we'd want a mechanism to apply the replacer to all snapshots. It gets verbose having to specify the replacer per use.

Another way I was thinking of approaching the problem in the long run was to expose some static "registry" / map of custom class types -> serializers to register with the class / or to override.

e.g.

```
class Serializer(AmberDataSerializer):
    custom_classes = { msgspec.Struct: some_serializer }

File: src/syrupy/matchers.py [hunk 2]
```diff
@@ -30,29 +34,30 @@ def path_type(
     *,
     types: Tuple["PropertyValueType", ...] = (),
     strict: bool = True,
-    regex: bool = False,
+    replacer: "Replacer" = lambda data, _: Repr(AmberDataSerializer.object_type(data)),
+    **_: Any,
```
noahnu, line 38: Is this to make it easy to extend the API?

src/syrupy/matchers.py
```diff
@@ -30,29 +34,30 @@ def path_type(
...
-        return path == pattern
-
     def path_type_matcher(
         *, data: "SerializableData", path: "PropertyPath"
     ) -> Optional["SerializableData"]:
         path_str = ".".join(str(p) for p, _ in path)
         if mapping:
             for pattern in mapping:
-                if _path_match(path_str, pattern):
+                matches = re.fullmatch(pattern, path_str)
```
noahnu, line 56: The old version that checked the "regex" flag supported literal string matches. Consider trying to match the object:

```python
dictionary = {
    "this.is.a.valid.key": True,
}
```

In the new version, you'd have to escape the `.` otherwise it'll get interpreted as regex and match any character.

poetry.lock
iamogbz: Ran `inv install` again, jic.

src/syrupy/extensions/amber/serializer.py
iamogbz: Changed the name to `object_as_named_tuple_class`.

> It gets verbose having to specify the replacer per use.

True, the `replacer` is a low hanging fruit that allows both functionality I wanted to add here. I like the `custom_classes` idea, it could even be used in the `__matcher` assertion property.

Another option would be to allow a change to the `_matcher` property that won't be reverted after each assertion.

```py
@pytest.fixture
def snapshot_bypass_my_custom_repr(snapshot):
  snapshot._matcher = path_type(
    types=(MyCustomReprClass,),
    replacer=lambda data, _: AmberDataSerializer.object_as_named_tuple(data),
  )
  return snapshot

def test_snapshot_object_as_named_tuple_class(snapshot_bypass_my_custom_repr):
    assert MyCustomReprClass() == snapshot_bypass_my_custom_repr(
        exclude=props("prop1"),
    )
```

src/syrupy/matchers.py
iamogbz: Yup and maintain backwards compatibility with the removal of the `regex` kwarg.

src/syrupy/matchers.py
iamogbz: True, also noticed there was no way to combine both literal string matches and regex matches in a single matcher even in the old version without escaping the literal string versions.

noahnu (approved):

PR #796, opened 2023-08-28
fix: remove colored dependency

noahnu:
## Description

Removes the last dependency from syrupy. We're now a zero-dependency project (improves compatibility).


src/syrupy/terminal.py
```diff
@@ -13,40 +9,52 @@ def _is_color_disabled() -> bool:
     return any(map(get_env_value, DISABLE_COLOR_ENV_VARS))
 
 
-def _attr(color: Any) -> str:
+@dataclass
+class TerminalCodes:
```
noahnu, line 13: This is mostly taken from the colored package: https://gitlab.com/dslackw/colored/-/blob/master/colored/colored.py?ref_type=heads.

PR #802, opened 2023-09-01
feat: support setting defaults

### LINKED ISSUE #800: Subsequent asserts seem to ignore the matcher
Hello! I've been recently exploring this tool and I wanted to create a fixture that I can re-use with my specific matcher, but it seems that subsequent calls in the same test ignore this matcher. I'm not sure if this is intended or not. I wasn't able to find a definitive answer. Apologies if this has been asked before!

**To reproduce**:

1. Run the following test with pytest:
```python
@pytest.fixture
def my_snapshot(snapshot):
    return snapshot(matcher=path_type({"field": (str,)}))


def test_snapshotting(my_snapshot):
    assert {"field": "string"} == my_snapshot
    assert {"field": "string"} == my_snapshot
```

This results in the following snapshot:
```
# serializer version: 1
# name: test_snapshotting
  dict({
    'field': str,
  })
# ---
# name: test_snapshotting.1
  dict({
    'field': 'string',
  })
# ---

```

**Expected behavior**

I would've expected that both snapshots would compare the type, instead of only the first one is saved with the type and the second one saves the value like the matcher wasn't there at all.

Providing the matcher to both asserts with `my_snapshot` solves this issue, but I don't find this copy-paste feasible too much in a codebase with over a thousand tests. A generic fixture that I define once and re-use would be ideal.

**Screenshots**

- OS: MacOS 13.4.1
- Syrupy Version: 4.4.0
- Python Version: 3.11

**Additional context**

michaljelonek:
## Description

- implements `with_defaults` that allows setting `include`, `exclude`, `matcher` and `extension_class`
- changes internal of `use_extension` to use `with_defaults`, functionally unchanged

## Related Issues

Closes #800, subsequent asserts seem to ignore the matcher.

## Checklist

## Additional Comments

Unsure what extra tests to add, but it seems like there would be benefit in more. Coverage seems to be at the same percentage.


noahnu (approved):
This is excellent! Thank you for the contribution!

noahnu:
@all-contributors add @michaljelonek for code

Issue #687, opened 2023-02-02
Multiple image comparison extension types in one test?


drjasonharrison-vp-eio:
I'm starting with the example test_custom_image_extension ( https://github.com/tophat/syrupy/blob/main/tests/examples/test_custom_image_extension.py ) and wondering how to extend this to multiple image formats. For example JPEG and PNG.

```
from syrupy.extensions.single_file import SingleFileSnapshotExtension

class JPEGImageExtension(SingleFileSnapshotExtension):
    @property
    def _file_extension(self) -> str:
        return "jpg"

class PNGImageExtension(SingleFileSnapshotExtension):
    @property
    def _file_extension(self) -> str:
        return "png"

@pytest.fixture
def snapshot_jpeg(snapshot):
    return snapshot.use_extension(JPEGImageExtension)

@pytest.fixture
def snapshot_png(snapshot):
    return snapshot.use_extension(PNGImageExtension)
```

and then pass both fixtures into the test:

```
def create_image(format):
   if format == 0:
        return "JPEG_IMAGE_DATA", "jpeg_image.jpg"
   
    return "PNG_IMAGE_DATA", "png_image.png"

def test_image_format(snapshot, snapshot_jpeg, snapshot_png):
    image_data, filename = create_image(0)
    filename_extension = filename.split(".")[-1]
    assert image_data == snapshot_jpeg(name=filename)

    image_data, filename = create_image(1)
    filename_extension = filename.split(".")[-1]
    assert image_data == snapshot_png(name=filename)
```

I'm not sure if a snapshot extension can take additional parameters, like the file extension (.jpg, .png) and select the inherited class of SingleFileSnapshotExtension. This would allow us to do something like:

```
def test_image_format(snapshot, snapshot_image):
    image_data, filename = create_image(0)
    filename_extension = filename.split(".")[-1]
    assert image_data == snapshot_image(name=filename, image_type=filename_extension)

    image_data, filename = create_image(1)
    filename_extension = filename.split(".")[-1]
    assert image_data == snapshot_image(name=filename, image_type=filename_extension)
```

noahnu:
```py
@pytest.fixture
def snapshot_image():
    def _function(image_type):
        if image_type == "jpeg":
            return snapshot.use_extension(JPEGImageExtension)
        if image_type == "png":
            return snapshot.use_extension(PNGImageExtension)
        raise ValueError("Bad image_type!")
    return _function


def test_image_format(snapshot, snapshot_image):
[...]

PR #814, opened 2023-09-25
feat: Support passing arbitrary arguments/context to custom extensions (Issue #700)

### LINKED ISSUE #700: Support passing arbitrary arguments/context to custom extensions

Would help with https://github.com/tophat/syrupy/issues/687. Not entirely sure the desired syntax. Perhaps something like:

```py
def test_case(snapshot):
    return "hello" == snapshot(kwargs={})
```

atharva-2001:
## Description

Related to https://github.com/tophat/syrupy/issues/687 and https://github.com/tophat/syrupy/issues/700.

Allowing custom `**kwargs` to custom extension classes.

e.g. This PR refactors the X submodule, applying Y's algorithm to improve Z by K percent.

## Related Issues

- Closes https://github.com/tophat/syrupy/issues/700. 

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -264,6 +264,7 @@ class SnapshotAssertion:
         extension_class: Optional[Type["AbstractSyrupyExtension"]] = None,
         matcher: Optional["PropertyMatcher"] = None,
         name: Optional["SnapshotIndex"] = None,
+        **kwargs: Any,
     ) -> "SnapshotAssertion":
         """
         Modifies assertion instance options
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -180,7 +180,7 @@ class SnapshotAssertion:
         include: Optional["PropertyFilter"] = None,
         matcher: Optional["PropertyMatcher"] = None,
         extension_class: Optional[Type["AbstractSyrupyExtension"]] = None,
-        extra_args: Optional[Dict] = None
+        extra_args: Optional[Dict] = None,
     ) -> "SnapshotAssertion":
         """
         Create new snapshot assertion fixture with provided values. This preserves
@@ -194,7 +194,7 @@ class SnapshotAssertion:
             test_location=self.test_location,
             extension_class=extension_class or self.extension_class,
             session=self.session,
-            extra_args=extra_args or self.extra_args
+            extra_args=extra_args or self.extra_args,
         )
 
     def use_extension(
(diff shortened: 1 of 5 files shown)


atharva-2001:
@noahnu I'm changing the classes which make up the `AbstractSyrupyExtension` class and also it's subclasses. Is this the right way to do this?

src/syrupy/assertion.py
```diff
@@ -264,6 +264,7 @@ def __call__(
         extension_class: Optional[Type["AbstractSyrupyExtension"]] = None,
         matcher: Optional["PropertyMatcher"] = None,
         name: Optional["SnapshotIndex"] = None,
+        **kwargs: Any,
```
noahnu, line 267: I'm a bit worried about name collisions by just forwarding arbitrary arguments, vs. something like:

```
assert "hi" == snapshot(extra={...})
```

I know it doesn't read as nicely, but you can always add syntactic sugar via a fixture:

```
@pytest.fixture
def snapshot(named_arg: str):
    return snapshot.with_defaults(extra={ "named_arg": named_arg })

def test_case(snapshot):
    assert "hi" == snapshot(named_arg="hi")
```

atharva-2001: I added the argument to both `__call__` and `with_defaults` method. I added it to the `__call__` argument because I wanted it to work with cases when I want to change the snapshot behaviour but don't want it to persist.

src/syrupy/assertion.py
```diff
@@ -280,6 +285,8 @@ def __call__(
             self.__with_prop("_custom_index", name)
         if diff is not None:
             self.__with_prop("_snapshot_diff", diff)
+        if extra_args:
+            self._extra_args = extra_args
```
noahnu, line 289: Use `self.__with_prop`. It does seam auto-cleanup

src/syrupy/assertion.py
```diff
@@ -300,6 +307,11 @@ def _assert(self, data: "SerializableData") -> bool:
         matches = False
         assertion_success = False
         assertion_exception = None
+        matcher_options = None
+        for key,value in self._extra_args.items():
+            if key == "matcher_options":
+                matcher_options = value
+                print("matcher_options", matcher_options)
```
noahnu, line 314: Remove the print

src/syrupy/assertion.py
```diff
@@ -300,6 +307,11 @@ def _assert(self, data: "SerializableData") -> bool:
         matches = False
         assertion_success = False
         assertion_exception = None
+        matcher_options = None
+        for key,value in self._extra_args.items():
+            if key == "matcher_options":
+                matcher_options = value
```
noahnu, line 313: This enforces a certain schema in the extra_args object which if we want to support arbitrary args we shouldn't do. Instead, we should propagate the "extra_args" to all relevant methods.

atharva-2001: Removed this. I also forwaded `extra_args` to different functions. Could you please tell me  what you think about this?

noahnu:
Changing the signatures of all these functions will break any existing extensions in the wild, unless they were already using kwargs. Even then, I spent a couple hours trying to pass the context (or extra_args as you called it) to all relevant functions and the PR grew a bit out of control... I think we need to move back to instance-based extensions (which was something we moved away from in syrupy v4).

I've put up https://github.com/tophat/syrupy/pull/816 to add "context", although context is only available in instance extension methods and there aren't many of those. Once that change is in, we'll have to start migrating from classmethods to instance methods. May need to do that in a syrupy v5 update since it'll be a breaking change to the API.

Going back to instance methods also plays nicely with some of the discussion in https://github.com/tophat/syrupy/pull/754

noahnu:
Closing as stale.

PR #823, opened 2023-10-19
feat(serializer): add support for FunctionType serialization

### LINKED ISSUE #808: More generic serialization of function types

When trying to compare snapshots with attributes that contain functions, the function repr is returned as
`<function request_response.<locals>.app at 0x7fbf103dc0e0>`

The "at 0x..." part changes with each run of the test, I'm assuming it's a memory location.

I'd like the `at 0x...` part to be removed.

I've worked around this using a matcher:
```
    def function_matcher(data, path):
        if isinstance(data, FunctionType):
            return f"<function '{data.__qualname__}'>"
        else:
            return data
```

**Additional context**

This could be implemented in the ambr and json snapshot extensions.

In ambr (serializer.py):

```
...
        elif isinstance(data, FunctionType):
            serialize_method = cls.serialize_function
...
        
    @classmethod
    def serialize_function(cls, data: FunctionType, *, depth: int = 0, **kwargs: Any) -> str:
        return cls.__serialize_plain(data=f"<{FunctionType.__name__} {data.__qualname__}>", depth=depth)
```

In json:

```
        if isinstance(data, FunctionType):
            return f"<{FunctionType.__name__} '{data.__qualname__}'>"
```

ManiacDC:
* add FunctionType serialization in amber and json extensions
* add tests

## Description

add FunctionType serialization in amber and json extensions

## Related Issues

- Closes #808 , More generic serialization of function types

## Checklist

## Additional Comments

I used inspect as I didn't want to re-invent the wheel. Is this satisfactory?

Also of minor concern - this changes the format of existing FunctionType serialization. I don't think this realistically causes a problem as the prior serialization would change every time it ran anyway.

### DIFF
--- a/src/syrupy/extensions/amber/serializer.py
+++ b/src/syrupy/extensions/amber/serializer.py
@@ -1,7 +1,9 @@
 import collections
+import inspect
 import os
 from collections import OrderedDict
 from types import (
+    FunctionType,
     GeneratorType,
     MappingProxyType,
 )
@@ -257,6 +259,8 @@ class AmberDataSerializer:
             serialize_method = cls.serialize_namedtuple
         elif isinstance(data, (list, tuple, GeneratorType)):
             serialize_method = cls.serialize_iterable
+        elif isinstance(data, FunctionType):
+            serialize_method = cls.serialize_function
         return serialize_method(**serialize_kwargs)
 
     @classmethod
@@ -335,6 +339,14 @@ class AmberDataSerializer:
             separator=": ",
             serialize_key=True,
             **kwargs,
+        )
+
+    @classmethod
+    def serialize_function(
+        cls, data: FunctionType, *, depth: int = 0, **kwargs: Any
+    ) -> str:
+        return cls.__serialize_plain(
+            data=f"{data.__qualname__}{str(inspect.signature(data))}", depth=depth
         )
 
     @classmethod
--- a/src/syrupy/extensions/json/__init__.py
+++ b/src/syrupy/extensions/json/__init__.py
@@ -1,7 +1,11 @@
 import datetime
+import inspect
 import json
 from collections import OrderedDict
-from types import GeneratorType
+from types import (
+    FunctionType,
+    GeneratorType,
+)
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -132,6 +136,12 @@ class JSONSnapshotExtension(SingleFileSnapshotExtension):
 
         if isinstance(data, (datetime.datetime,)):
             return data.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
+
+        if isinstance(data, FunctionType):
+            return (
+                f"<{FunctionType.__name__} "
+                f"'{data.__qualname__}{str(inspect.signature(data))}'>"
+            )
 
         if data.__class__.__repr__ != object.__repr__:
             return repr(data)
--- a/src/syrupy/extensions/json/__init__.py
+++ b/src/syrupy/extensions/json/__init__.py
@@ -140,7 +140,7 @@ class JSONSnapshotExtension(SingleFileSnapshotExtension):
         if isinstance(data, FunctionType):
             return (
                 f"<{FunctionType.__name__} "
-                f"'{data.__qualname__}{str(inspect.signature(data))}'>"
+                f"{data.__qualname__}{str(inspect.signature(data))}>"
             )
 
         if data.__class__.__repr__ != object.__repr__:
--- a/tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
+++ b/tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
@@ -217,6 +217,12 @@
 # name: test_empty_snapshot.1
   ''
 # ---
+# name: test_function_in_file
+  "function_to_test(var1, var2='test_val', var3: str = 'test_val2', *, kwvar1, kwvar2='some_val') -> str"
+# ---
+# name: test_function_local
+  "test_function_local.<locals>.local_function_to_test(var1, var2='test_val', var3: str = 'test_val2', *, kwvar1, kwvar2='some_val') -> int"
+# ---
 # name: test_list[actual0]
   list([
   ])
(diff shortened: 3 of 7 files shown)


ManiacDC:
@noahnu any idea why the Benchmark workflow is failing? Doesn't seem to be an issue with the PR changes.

noahnu:
You can ignore the benchmark. It always fails for external contributors. There are credentials that don't get exposed.

I'll try to review this pull request soon, thank you for the contribution!

tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
```diff
@@ -217,6 +217,12 @@
 # name: test_empty_snapshot.1
   ''
 # ---
+# name: test_function_in_file
+  "function_to_test(var1, var2='test_val', var3: str = 'test_val2', *, kwvar1, kwvar2='some_val') -> str"
```
noahnu, line 221: Is this going based off source code when using `inspect.signature`? I find it odd that we're getting the whitespace inconsistencies from the source code in the snapshot. It makes no functional difference and so I would expect to see it.

tests/syrupy/extensions/amber/test_amber_serializer.py
```diff
@@ -233,3 +233,22 @@ def test_ordered_dict(snapshot):
...
+
+
+def function_to_test(
+    var1, var2="test_val", var3: str = "test_val2", *, kwvar1, kwvar2="some_val"
+) -> str:
+    return "2"
+
+
+def test_function_in_file(snapshot):
+    assert snapshot() == function_to_test
```
noahnu, line 245:
```suggestion
    assert snapshot == function_to_test
```

no need to invoke snapshot

src/syrupy/extensions/json/__init__.py
```diff
@@ -133,6 +137,12 @@ def _filter(
         if isinstance(data, (datetime.datetime,)):
             return data.strftime("%Y-%m-%dT%H:%M:%S.%f%z")
 
+        if isinstance(data, FunctionType):
+            return (
+                f"<{FunctionType.__name__} "
+                f"'{data.__qualname__}{str(inspect.signature(data))}'>"
```
iamogbz, line 143: Serialization nitpick
```suggestion
                f"{data.__qualname__}{str(inspect.signature(data))}>"
```
Result
```json
"<function function_to_test(var1, var2='test_val', var3: str = 'test_val2', *, kwvar1, kwvar2='some_val') -> str>"
```

tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
iamogbz: Does the resulting signature string differ on different python/systems? Might just not be an issue. Though an option is removing/normalising whitespace to make it more interoperable? (like was done with  NL/CR)

iamogbz (commented):
LGTM

tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
ManiacDC: So that's actually how inspect.signature() formats the output. It puts spaces around the '=' sign if following a typehint. If I remove the spaces in the source code, it adds them back in. If I add a typehint to a different variable, it adds the spaces in.

I could strip the whitespace, but since the behavior is consistent I'm inclined to leave it. Thoughts?

ManiacDC:
Thanks for the review! I've updated with the suggested changes.

tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
ManiacDC: @iamogbz the result of the signature string should not differ on different systems, the output formatting is explicit. However, on different python versions it could, if they change how signature works. Does Syrupy guarantee formatting across python versions? If so, I'd need to write my own formatter.

src/syrupy/extensions/amber/serializer.py
```diff
@@ -337,6 +341,14 @@ def serialize_dict(
             **kwargs,
         )
 
+    @classmethod
+    def serialize_function(
+        cls, data: FunctionType, *, depth: int = 0, **kwargs: Any
+    ) -> str:
+        return cls.__serialize_plain(
+            data=f"{data.__qualname__}{str(inspect.signature(data))}", depth=depth
```
ManiacDC, line 349: Looking at the doc for inspect.signature again, I'm wondering if I should set `eval_str=False`. @noahnu , thoughts?

https://docs.python.org/3/library/inspect.html#inspect.signature

tests/syrupy/extensions/amber/__snapshots__/test_amber_serializer.ambr
noahnu: > Does Syrupy guarantee formatting across python versions?

We provide no explicit guarantee and I think there's an expectation that snapshots _may_ differ when you update underlying libraries/tech. That being said, we run our test suites against multiple versions of python and operating systems (though not multiple implementations) like cython etc.

Regarding whitespace, as long as it's consistent, I think it's fine.

src/syrupy/extensions/amber/serializer.py
noahnu: It defaults to false, so no action required.

noahnu:
Snapshot needs to be updated.

src/syrupy/extensions/amber/serializer.py
ManiacDC: Oh, right. lol.

ManiacDC:
> Snapshot needs to be updated.

Yeah I had left some testing code in the amber file by accident, fixed. Thanks!

ManiacDC:
Caught the formatting issue, but not fast enough. Sorry about that, used to format on save and I haven't setup vscode for this repo yet.

noahnu (approved):

noahnu:
@all-contributors-bot add @ManiacDC for code

noahnu:
@all-contributors add @ManiacDC for code

ManiacDC:
Thank you!

Issue #597, opened 2022-04-13
Summary of updated snapshots


KaeganCasey:
**Is your feature request related to a problem? Please describe.**

When I run snapshot update and only a few snapshots have changed or a few new ones get created it currently gives a summary saying "x snapshots have been updated". 

**Describe the solution you'd like**

It would be wonderful if it could print a complete list of the names of each of the snapshots so that you can feel more confident that only the ones you expected to be updated / created were changed. This would give me more confidence when updating snapshots and a clear list of the ones I should go look at manually to validate their correctness. Ideally it would have a separate section where it prints the names of any snapshots that were updated and then another section for printing the names of newly created snapshots.

Thank you very much for this awesome library it saves me so much time and effort and has beautiful syntax.

noahnu:
1. Perhaps we can extend `--snapshot-details`. This should be relatively straightforward if you're interested in contributing (otherwise no ETA on my side at the moment, I'm trying to get some more internal contributors to work through the open issues list).
2. I didn't realize it was via a Pycharm plugin. I think that should be doable. I'll need to dig into the plugin some more. My main concern with repr is that we'd have to perform an IO operation to read the snapshot.

KaeganCasey:
That all sounds awesome! Thank you for responding so quickly! No rush as these are mostly cosmetic enhancements.

noahnu:
2 is implemented in https://github.com/syrupy-project/syrupy/releases/tag/v4.7.0. Will edit his issue so it's asking 1 question

noahnu:
This should be a simple change to implement. The reporter prints out a summary for creates/updates here: https://github.com/syrupy-project/syrupy/blob/8c1208d7dbff2c9de2a559f5e6a9783bb12a629a/src/syrupy/report.py#L328. All we'd have to do is check `if self.include_snapshot_details:` and then print some additional info. All the required context is already available.

Open to contributions on this issue.

noahnu:
Included in the v5 release. Might end up backporting it.

noahnu:
Backported to [4.9.0](https://github.com/syrupy-project/syrupy/releases/tag/v4.9.0)

PR #864, opened 2024-06-15
feat: Added a new CLI flag: --snapshot-patch-pycharm-diff

### LINKED ISSUE #675: PyCharm diff comparison failure doesn't work with custom classes
**Describe the bug**

When comparing a failed snapshot test on an object without a custom `__repr__` (i.e. relies on syrupy's serializer), PyCharm incorrectly displays the actual output as the default `__repr__` (e.g. `...Mansion object at 0xBEEF2020`) instead of what syrupy serialises it to (e.g. `Mansion(rooms=[Room(...)])`).

This might be a PyCharm bug, or Syrupy or somehow both. Just thought I'd make this issue as it tripped me up at work for over an hour and I imagine a fair few people use PyCharm and syrupy together.

**To reproduce**

1. checkout [this repo I made](https://github.com/lindenwells/syrupy-pycharm-diff-failure-example)
2. create a virtualenv within and `pip install syrupy==3.0.2`
3. find the singular test within the run configurations and notice that it fails.
4. in the failed test output, scroll to the clickable '\<Click to see difference\>' text: image
5. click that text and you should get a Comparison Failure interface like this: image

**Expected behavior**

This is roughly what I should see in the Comparison Failure window.
image
I know this is wrong because when I run `pytest --snapshot-update`, it updates the snapshot like this. 

**Screenshots**
added throughout where appropriate

**Environment:**

- OS: Windows 11 Pro 21H2
- Syrupy Version: 3.0.2
- Python Version: 3.9.13
- PyCharm Version: 2022.3 (Community Edition), Build #PC-223.7571.203, built on December 1, 2022, Runtime version: 17.0.5+1-b653.14 amd64, VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.

**Additional context**

This is similar to https://github.com/tophat/syrupy/issues/572 but not the same.

UltimateLobster:
(Only relevant for Pycharm users) If the flag is passed, Syrupy will import and extend Pycharm's default diff script to provide more meaning full diffs when viewing snapshots mismatches in Pycharm's diff viewer

## Description

When running snapshots tests in Pycharm (in fact, when running tests in general in Pycharm), it has this cool feature where you can see the difference between the `expected` and the `actual` objects in a diff viewer:

```python
def test_example():
    expected = {
        "a": 1,
        "b": 2,
        "c": 3,
    }
    actual = {
        "a": 1,
        "b": 2,
        "c": 5,
    }
    assert actual== expected
```
After running `test_example`, and looking at Pycharm's console, we can see a link:
image

By clicking the `<Click to see difference>` link we can easily see the difference and know what the problem was:
image

However, lets run some snapshot tests this time:
```python
@pytest.fixture
def json_snapshot(snapshot):
    return snapshot.with_defaults(extension_class=JSONSnapshotExtension)


def test_example(json_snapshot):
    actual = {
        "a": 1,
        "b": 2,
        "c": 5,
    }
    assert actual == json_snapshot
```

This time, Pycharm will show an incorrect difference that is hard to parse out:
image

This happens because Pycharm only sees an `actual` and an `expected` object, it doesn't have a way to know that the `actual` objects needs further parsing before the diff calculation.

This PR adds a new flag: --snapshot-patch-pycharm-diff. If the flag is active, Syrupy will attempt to import Pycharm's built-in `teamcity-messages` package (which Pycharm automatically adds to PYTHONPATH when running scripts in order to make it importable). If the import was successful, the `teamcity.diff_tools.EqualsAssertionError` object (which is responsible to calculating the string representation of the `expected` and `actual` objects in the diff view) will be patched. 

The patched object will behave the same as before. However, if either the `expected` object or the `actual` objects are instances of `syrupy.assertions.SnapshotAssertion`, the `expected` and `actual` objects will be replaced by the snapshot's recalled data and asserted data respectively.

Now, if we activate the new flag, the diff will be meaningful again:
image

## Related Issues

This PR should resolve #675 as well as the second part of #597 

## Checklist

- [X ] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary).

## Additional Comments

As you might see in the checklist, there are a few missing parts in this PR. This is because I wasn't sure how to go about certain parts of the checklist.

1. At first I wanted to put the entire documentation in the README.md file (because that's where I saw the rest of the configuration going to. However, I wanted to put some images to have a more clear explanation and I wasn't sure where would be a good place to put them (if it is ok to put any at all).
2. I have no idea how should I go about testing this feature considering its results are accessible via Pycharm's UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,12 +132,13 @@ Both options will generate equivalent snapshots but the latter is only viable wh
 These are the cli options exposed to `pytest` by the plugin.
 
 | Option                         | Description                                                                                                                    | Default                                                                                                      |
-| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------ |
+| ------------------------------ |--------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------|
 | `--snapshot-update`            | Snapshots will be updated to match assertions and unused snapshots will be deleted.                                            | `False`                                                                                                      |
 | `--snapshot-details`           | Includes details of unused snapshots (test name and snapshot location) in the final report.                                    | `False`                                                                                                      |
 | `--snapshot-warn-unused`       | Prints a warning on unused snapshots rather than fail the test suite.                                                          | `False`                                                                                                      |
 | `--snapshot-default-extension` | Use to change the default snapshot extension class.                                                                            | [AmberSnapshotExtension](https://github.com/syrupy-project/syrupy/blob/main/src/syrupy/extensions/amber/__init__.py) |
 | `--snapshot-no-colors`         | Disable test results output highlighting. Equivalent to setting the environment variables `ANSI_COLORS_DISABLED` or `NO_COLOR` | Disabled by default if not in terminal.                                                                      |
+| `--snapshot-patch-pycharm-diff`| Override Pycharm's default diffs viewer when looking at snapshot diffs. More information in [Using Syrupy from Pycharm]        | `False`                                                                                                      |
 
 ### Assertion Options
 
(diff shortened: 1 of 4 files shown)


lindenwells:
Cool idea! You did what I wanted to but didn't have time to. You're right tho, it's kinda messy. Would this logic be better suited to living in PyCharm or does it absolutely need to be on the syrupy side?

UltimateLobster:
> Cool idea! You did what I wanted to but didn't have time to. You're right tho, it's kinda messy. Would this logic be better suited to living in PyCharm or does it absolutely need to be on the syrupy side?

I have no idea how Jetbrains manages their scripts, but the script in question is just a class that generically receives an "expected" and "actual" objects and formats them. It doubt Jetbrains would rush into adding extra code handling specifically for objects of a single library. Unless they offically add support for custom diff calculations (like a `__calculate_diff__` or something to check on every object), I would assume the change would have to be strictly on Syrupy side.

Regardless, I guess it wouldn't hurt to ask around and check their stance on this issue.

noahnu:
I've added test coverage and will get this merged. Thank you for the contribution!

noahnu:
@all-contributors add @UltimateLobster for feature

noahnu:
https://github.com/all-contributors add @UltimateLobster for code

noahnu:
@all-contributors add @UltimateLobster for code

PR #892, opened 2024-09-25
fix: allow snapshot dir to be different

### LINKED ISSUE #868: Overwriting dirname causes test suite not to fail if snapshots unused
**Describe the bug**

Within Home Assistant we have a SnapshotExtension to overwrite the default `__snapshots__` dir name with `snapshots`. But today I actually noticed that this had the side effect that it doesn't fail the test suite if there are unused snapshots in the snapshots folder.

```py
class HomeAssistantSnapshotExtension(AmberSnapshotExtension):
    """Home Assistant extension for Syrupy."""

    VERSION = "1"
    """Current version of serialization format.

    Need to be bumped when we change the HomeAssistantSnapshotSerializer.
    """

    serializer_class: type[AmberDataSerializer] = HomeAssistantSnapshotSerializer

    @classmethod
    def dirname(cls, *, test_location: PyTestLocation) -> str:
        """Return the directory for the snapshot files.

        Syrupy, by default, uses the `__snapshosts__` directory in the same
        folder as the test file. For Home Assistant, this is changed to just
        `snapshots` in the same folder as the test file, to match our `fixtures`
        folder structure.
        """
        test_dir = Path(test_location.filepath).parent
        return str(test_dir.joinpath("snapshots"))

```

**To reproduce**
Add the SnapshotExtension. Have it generate a snapshot. Add a fake entry to the snapshot. Run the test suite.

**Expected behavior**

I'd expect it to fail the test suite because there are entries that are not found. I can imagine that we have quite a basic way of overwriting the directory and can imagine people with way more interesting setups for the snapshots which would probably break with the expected behaviour. So I would not be surprised if the failure is turned off when the dirname is overwritten, but I'd then at least expect to read something in a docstring or in the docs.

**Screenshots**

- OS: mac OS
- Syrupy Version: 4.6.1
- Python Version: 3.12

**Additional context**

joostlek:
## Description

This PR aims to allow setting the dirname to something else than `__snapshots__` while maintaining the functionality of a failing test suite when snapshots are unused.

## Related Issues

- Closes #868, Where overwriting a dirname causes a test suite not to fail.

## Checklist

## Additional Comments

I think I changed the `test_walk_dir_skips_non_snapshot_path` test in a wrong way, but since the function this is testing (`walk_snapshot_dir`) is always called in context, this isn't a real use case afaict

### DIFF
--- a/src/syrupy/utils.py
+++ b/src/syrupy/utils.py
@@ -37,8 +37,6 @@ def in_snapshot_dir(path: Path) -> bool:
 
 def walk_snapshot_dir(root: str) -> Iterator[str]:
     for filepath in Path(root).rglob("*"):
-        if not in_snapshot_dir(filepath):
-            continue
         if not filepath.name.startswith(".") and filepath.is_file():
             yield str(filepath)
 
--- a/tests/integration/test_snapshot_outside_directory.py
+++ b/tests/integration/test_snapshot_outside_directory.py
@@ -1,9 +1,9 @@
 import pytest
 
 
-@pytest.fixture
-def testcases(testdir, tmp_path):
-    dirname = tmp_path.joinpath("__snapshots__")
+@pytest.fixture(params=["__snapshots__", "snapshots"])
+def testcases(testdir, tmp_path, request):
+    dirname = tmp_path.joinpath(request.param)
     testdir.makeconftest(
         f"""
         import pytest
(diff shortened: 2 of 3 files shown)


noahnu:
> I think I changed the test_walk_dir_skips_non_snapshot_path test in a wrong way, but since the function this is testing (walk_snapshot_dir) is always called in context, this isn't a real use case afaict

Yeah that seems alright to me. I confirmed walk_snapshot_dir is only ever called with the snapshot dirname. I think it was called with other directories in some previous iteration of the codebase.

noahnu:
@all-contributors add @joostlek for code

joostlek:
This is awesome, thank you so much

PR #606, opened 2022-05-28
wip: defer snapshot writing

noahnu:
## Description

e.g. This PR refactors the X submodule, applying Y's algorithm to improve Z by K percent.

## Related Issues

- Closes #0, a short description of the linked issue.

## Checklist

- [ ] This PR has sufficient documentation.
- [ ] This PR has sufficient test coverage.
- [ ] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary).

## Additional Comments

No additional comments.

PR #927, opened 2024-12-23
fix: check current session's pending-write queue when recalling snapshots (e.g. diffing)

### LINKED ISSUE #926: Diff snapshot requires two test runs for new tests and two to 'stabilise' after changes to base snapshot
**Describe the bug**

Currently, using `diff` seems to take two test runs with `--snapshot-update` to have the diff computed correctly in two cases:

- if a test is newly added with both a base snapshot and a diff, the diff snapshot will fail with a `SnapshotDoesNotExist` error on the first run, but work on the second
- if the "base"/original snapshot has a change, the first test run will have those changes appear in the diff, and a second run is required to clear them out

**To reproduce**

For a new test:

1. Create a new test like the following:
    ```python
    base = {"A": 1}
    def test_foo(snapshot):
        assert {**base} == snapshot(name="a")
        assert {**base, "B": 2} == snapshot(name="b", diff="a")
    ```
2. run `pytest --snapshot-update`
3. BUG: observe it fails with an error like:
    ```
    >       assert {**base, "B": 2} == snapshot(name="b", diff="a")
    E       assert [+ received] == [- snapshot]
    E         syrupy.exceptions.SnapshotDoesNotExist
    E         Traceback (most recent call last):
    E           File "/private/var/folders/sv/vd266m4d4lvctgs2wpnhjs9w0000gn/T/tmp.Kn7Y7pGLWR/.venv/lib/python3.10/site-packages/syrupy/assertion.py", line 322, in _assert
    E             raise SnapshotDoesNotExist()
    
    test_foo.py:4: AssertionError
    ```
4. run `pytest --snapshot-update` again (no other changes) and observe it works

For an existing test:

5. Take the test above (after running `pytest --snapshot-update` twice) and edit `base` to have an extra key, `base = {"A": 1, "C": 3}`
6. Run `pytest --snapshot-update`
7. BUG: observe the `__snapshot__/test_foo.ambr` file contains the following, and, in particular, the diff says that `'C': 3` was added... but it appears in the `a` snapshot too, i.e. appears in both snapshots and thus shouldn't be in the diff!
    ```
    # serializer version: 1
    # name: test_foo[a]
      dict({
        'A': 1,
        'C': 3,
      })
    # ---
    # name: test_foo[b]
          ...
      +   'B': 2,
      +   'C': 3,
        ...
    # ---
    ```
8. Run `pytest` (no `--snapshot-update`) and observe failures:
    ```
    >       assert {**base, "B": 2} == snapshot(name="b", diff="a")
    E       AssertionError: assert [+ received] == [- snapshot]
    E               ...
    E           +   'B': 2,
    E         - +   'C': 3,
    E             ...
    
    test_foo.py:4: AssertionError
    ```

An all-in-one shell script reproducer: save it and run with `bash ./script.sh`

<details>

```shell
cd $(mktemp -d)

python -m venv .venv
. .venv/bin/activate

pip install syrupy==4.8.0 colored==1.4.4 exceptiongroup==1.2.2 iniconfig==2.0.0 packaging==24.2 pluggy==1.5.0 pytest==7.4.4 tomli==2.2.1

cat > test_foo.py <<EOF
base = {"A": 1}
def test_foo(snapshot):
    assert {**base} == snapshot(name="a")
    assert {**base, "B": 2} == snapshot(name="b", diff="a")
EOF

echo '**** BUG: first run fails with `raise SnapshotDoesNotExist`'
pytest test_foo.py --snapshot-update
echo '**** OKAY: second run passes'
pytest test_foo.py --snapshot-update

# add a new field C to base
cat > test_foo.py <<EOF
base = {"A": 1, "C": 3}
def test_foo(snapshot):
    assert {**base} == snapshot(name="a")
    assert {**base, "B": 2} == snapshot(name="b", diff="a")
EOF

echo '**** BUG: running after a change has the diff snapshot think C is added (should only be B)'
pytest test_foo.py --snapshot-update
cat __snapshots__/test_foo.ambr
echo '**** ... which means a second non-snapshot test run fails'
pytest test_foo.py
```

</details>

**Expected behavior**

Running `pytest --snapshot-update` should make the complete set of updates required, i.e. assuming the snapshots are reproducible, the both:

- running `pytest --snapshot-update` a second time should make no changes
- running `pytest` alone should pass

**Screenshots**

- OS: macOS 15
- Syrupy Version: 4.8
- Python Version: 3.10

**Additional context**

Maybe very slightly related to https://github.com/syrupy-project/syrupy/issues/608

(Thanks for syrupy! :pancakes: )

huonw:
## Description

This PR fixes `assert ... == snapshot(diff=...)` mode to work better in two scenarios:

- Adding two snapshots at the same time (i.e. not already "on disk"), where the second snapshot is a diff from the first
- Changing the base data of a diff'd snapshot

In particular, it seems like the nice performance optimisation of #606, meant the "recalling" used for diffing wasn't finding newly created or updated data, and instead just reading whatever snapshot is available on disk.

This PR resolves the problem by having the recalling managed via the `SnapshotSession`, which can thus check its write queue before reading disk.

## Related Issues

- Closes #926 

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/assertion.py
+++ b/src/syrupy/assertion.py
@@ -377,11 +377,7 @@ class SnapshotAssertion:
     ) -> Tuple[Optional["SerializableData"], bool]:
         try:
             return (
-                self.extension.read_snapshot(
-                    test_location=self.test_location,
-                    index=index,
-                    session_id=str(id(self.session)),
-                ),
+                self.session.recall_snapshot(self.extension, self.test_location, index),
                 False,
             )
         except SnapshotDoesNotExist:
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -13,7 +13,7 @@ import pytest
 from syrupy.constants import PYTEST_NODE_SEP
 
 
-@dataclass
+@dataclass(frozen=True)
 class PyTestLocation:
     item: "pytest.Item"
     nodename: Optional[str] = field(init=False)
@@ -23,27 +23,41 @@ class PyTestLocation:
     filepath: str = field(init=False)
 
     def __post_init__(self) -> None:
+        # NB. we're in a frozen dataclass, but need to transform the values that the caller
+        # supplied... we do so by (ab)using object.__setattr__ to forcibly set the attributes. (See
+        # rejected PEP-0712 for an example of a better way to handle this.)
+        #
+        # This is safe because this all happens during initialization: `self` hasn't been hashed
+        # (or, e.g., stored in a dict), so the mutation won't be noticed.
         if self.is_doctest:
             return self.__attrs_post_init_doc__()
         self.__attrs_post_init_def__()
 
     def __attrs_post_init_def__(self) -> None:
         node_path: Path = getattr(self.item, "path")  # noqa: B009
-        self.filepath = str(node_path.absolute())
+        # See __post_init__ for discussion of object.__setattr__
+        object.__setattr__(self, "filepath", str(node_path.absolute()))
         obj = getattr(self.item, "obj")  # noqa: B009
-        self.modulename = obj.__module__
-        self.methodname = obj.__name__
-        self.nodename = getattr(self.item, "name", None)
-        self.testname = self.nodename or self.methodname
+        object.__setattr__(self, "modulename", obj.__module__)
+        object.__setattr__(self, "methodname", obj.__name__)
+        object.__setattr__(self, "nodename", getattr(self.item, "name", None))
+        object.__setattr__(self, "testname", self.nodename or self.methodname)
 
     def __attrs_post_init_doc__(self) -> None:
         doctest = getattr(self.item, "dtest")  # noqa: B009
-        self.filepath = doctest.filename
+        # See __post_init__ for discussion of object.__setattr__
+        object.__setattr__(self, "filepath", doctest.filename)
         test_relfile, test_node = self.nodeid.split(PYTEST_NODE_SEP)
         test_relpath = Path(test_relfile)
-        self.modulename = ".".join([*test_relpath.parent.parts, test_relpath.stem])
-        self.nodename = test_node.replace(f"{self.modulename}.", "")
-        self.testname = self.nodename or self.methodname
+        object.__setattr__(
+            self,
+            "modulename",
+            ".".join([*test_relpath.parent.parts, test_relpath.stem]),
+        )
+        object.__setattr__(
+            self, "nodename", test_node.replace(f"{self.modulename}.", "")
+        )
+        object.__setattr__(self, "testname", self.nodename or self.methodname)
 
     @property
     def classname(self) -> Optional[str]:
--- a/src/syrupy/location.py
+++ b/src/syrupy/location.py
@@ -54,6 +54,7 @@ class PyTestLocation:
             "modulename",
             ".".join([*test_relpath.parent.parts, test_relpath.stem]),
         )
+        object.__setattr__(self, "methodname", None)
         object.__setattr__(
             self, "nodename", test_node.replace(f"{self.modulename}.", "")
         )
--- a/src/syrupy/session.py
+++ b/src/syrupy/session.py
@@ -64,7 +64,7 @@ class SnapshotSession:
 
     _queued_snapshot_writes: Dict[
         Tuple[Type["AbstractSyrupyExtension"], str],
-        List[Tuple["SerializedData", "PyTestLocation", "SnapshotIndex"]],
+        Dict[Tuple["PyTestLocation", "SnapshotIndex"], "SerializedData"],
     ] = field(default_factory=dict)
 
     def _snapshot_write_queue_key(
@@ -86,8 +86,8 @@ class SnapshotSession:
         index: "SnapshotIndex",
     ) -> None:
         key = self._snapshot_write_queue_key(extension, test_location, index)
-        queue = self._queued_snapshot_writes.get(key, [])
-        queue.append((data, test_location, index))
+        queue = self._queued_snapshot_writes.get(key, {})
+        queue[(test_location, index)] = data
         self._queued_snapshot_writes[key] = queue
 
     def flush_snapshot_write_queue(self) -> None:
@@ -97,7 +97,11 @@ class SnapshotSession:
         ), queued_write in self._queued_snapshot_writes.items():
             if queued_write:
                 extension_class.write_snapshot(
-                    snapshot_location=snapshot_location, snapshots=queued_write
+                    snapshot_location=snapshot_location,
+                    snapshots=[
+                        (data, loc, index)
+                        for (loc, index), data in queued_write.items()
+                    ],
                 )
         self._queued_snapshot_writes = {}
 
@@ -110,12 +114,10 @@ class SnapshotSession:
         """Find the current value of the snapshot, for this session, either a pending write or the actual snapshot."""
 
         key = self._snapshot_write_queue_key(extension, test_location, index)
-        queue = self._queued_snapshot_writes.get(key)
-        if queue:
-            # find the last (i.e. most recent) write to this index/location in the queue:
-            for queue_data, queue_test_location, queue_index in reversed(queue):
-                if queue_index == index and queue_test_location == test_location:
-                    return queue_data
+        queue = self._queued_snapshot_writes.get(key, {})
+        data = queue.get((test_location, index))
+        if data is not None:
+            return data
 
         # no queue, or no matching write, so just read the snapshot directly:
         return extension.read_snapshot(
--- a/src/syrupy/session.py
+++ b/src/syrupy/session.py
@@ -66,10 +66,10 @@ class SnapshotSession:
         default_factory=lambda: defaultdict(set)
     )
 
-    _queued_snapshot_writes: Dict[
+    _queued_snapshot_writes: DefaultDict[
         _QueuedWriteExtensionKey,
         Dict[_QueuedWriteTestLocationKey, "SerializedData"],
-    ] = field(default_factory=dict)
+    ] = field(default_factory=lambda: defaultdict(dict))
 
     def _snapshot_write_queue_keys(
         self,
@@ -92,9 +92,7 @@ class SnapshotSession:
         ext_key, loc_key = self._snapshot_write_queue_keys(
             extension, test_location, index
         )
-        queue = self._queued_snapshot_writes.get(ext_key, {})
-        queue[loc_key] = data
-        self._queued_snapshot_writes[ext_key] = queue
+        self._queued_snapshot_writes[ext_key][loc_key] = data
 
     def flush_snapshot_write_queue(self) -> None:
         for (
@@ -109,7 +107,7 @@ class SnapshotSession:
                         for (loc, index), data in queued_write.items()
                     ],
                 )
-        self._queued_snapshot_writes = {}
+        self._queued_snapshot_writes.clear()
 
     def recall_snapshot(
         self,
@@ -122,8 +120,7 @@ class SnapshotSession:
         ext_key, loc_key = self._snapshot_write_queue_keys(
             extension, test_location, index
         )
-        queue = self._queued_snapshot_writes.get(ext_key, {})
-        data = queue.get(loc_key)
+        data = self._queued_snapshot_writes[ext_key].get(loc_key)
         if data is not None:
             return data
 
--- a/src/syrupy/session.py
+++ b/src/syrupy/session.py
@@ -66,6 +66,13 @@ class SnapshotSession:
         default_factory=lambda: defaultdict(set)
     )
 
+    # For performance, we buffer snapshot writes in memory before flushing them to disk. In
+    # particular, we want to be able to write to a file on disk only once, rather than having to
+    # repeatedly rewrite it.
+    #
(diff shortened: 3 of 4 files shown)


src/syrupy/session.py
```diff
@@ -93,6 +101,27 @@ def flush_snapshot_write_queue(self) -> None:
...
+    ) -> Optional["SerializedData"]:
+        """Find the current value of the snapshot, for this session, either a pending write or the actual snapshot."""
+
+        key = self._snapshot_write_queue_key(extension, test_location, index)
+        queue = self._queued_snapshot_writes.get(key)
+        if queue:
+            # find the last (i.e. most recent) write to this index/location in the queue:
+            for queue_data, queue_test_location, queue_index in reversed(queue):
+                if queue_index == index and queue_test_location == test_location:
+                    return queue_data
```
huonw, line 118: This loops is a potential performance issue. In particular: it scales like O(length of `queue`), i.e. number of pending writes.

This function is called twice in `SnapshotAssertion._assert`:

- unconditionally: https://github.com/syrupy-project/syrupy/blob/ef8189c68460593fead9c484405b755e272c8cca/src/syrupy/assertion.py#L316
- conditionally, only when diffing: https://github.com/syrupy-project/syrupy/blob/ef8189c68460593fead9c484405b755e272c8cca/src/syrupy/assertion.py#L320

If every assertion in a ambr file needs an update, this leads to quadratic O(number of assertions in that file^2) behaviour:

- the first `assert ... == snapshot` has an empty `queue`, and adds an item
- the second checks one element of the queue (doesn't match), and adds an item
- the third checks two elements of the queue (don't match) and adds an item
- ...
- the Nth checks N - 1 elements of the queue (all don't match) and adds an item

i.e. N assertions ends up iterating over 0 + 1 + 2 + ... + (N - 1) = (N - 1) * (N - 2) / 2 = O(N^2) elements of `queue`.

I think this is probably okay:

- this scales only per snapshot location, not across the full testing session, since each `queue` is just for one key, which incorporates the snapshot location (i.e. if one has 10 assertions in each of 100 test files, this is scaling with N = 10, not N = 10 * 100).
- how often does one have to completely re-write a _whole_ file?
- the work here is individually cheap (i.e. the constant factor is low)

src/syrupy/assertion.py
```diff
@@ -377,11 +377,7 @@ def _recall_data(
     ) -> Tuple[Optional["SerializableData"], bool]:
         try:
             return (
-                self.extension.read_snapshot(
-                    test_location=self.test_location,
-                    index=index,
-                    session_id=str(id(self.session)),
```
huonw, line 383: I note that #606 added this `session_id` parameter, which I imagine was meant to help with circumstances like this, but it seems like it ends up ignored/unused in practice?

- Amber doesn't use `cache_key`: https://github.com/tophat/syrupy/blob/ef8189c68460593fead9c484405b755e272c8cca/src/syrupy/extensions/amber/__init__.py#L57-L66
- Single-file doesn't use `session_id`: https://github.com/tophat/syrupy/blob/ef8189c68460593fead9c484405b755e272c8cca/src/syrupy/extensions/single_file.py#L92-L103

And, in practice, I think it may be quite hard to use, since the relevant cached data is held in memory in the session itself (i.e. the extension `_read_snapshot_data_from_location` method calls don't really have access to it at all).

huonw (commented):
As always, thank you for Syrupy!

src/syrupy/assertion.py
noahnu: The session_id is used as the cache_key argument in Amber's `__cacheable_read_snapshot` method. Although it's not used in the function, it's used by the lru_cache decorator (which cases based on the kwargs I believe). So it essentially invalidates the cache

noahnu:
# Benchmark

<details>

| Benchmark suite | Current: dbf395a21ad11f1bd5c9d4ef7375c0b7a8ca7d43 | Previous: d283e87c2b8d1d5623ea3258653d579757bfa78c | Ratio |
|-|-|-|-|
| `benchmarks/test_1000x.py::test_1000x_reads` | `0.7313520234418077` iter/sec (`stddev: 0.06327692389499665`) | `0.8206510272360953` iter/sec (`stddev: 0.06414211339510058`) | `1.12` |
| `benchmarks/test_1000x.py::test_1000x_writes` | `0.5937400226479332` iter/sec (`stddev: 0.16220190601651813`) | `0.7296660677418227` iter/sec (`stddev: 0.160353318333188`) | `1.23` |
| `benchmarks/test_standard.py::test_standard` | `0.6769499288981089` iter/sec (`stddev: 0.05068615977797293`) | `0.758693423573993` iter/sec (`stddev: 0.06688161104621605`) | `1.12` |

</details>

This comment was automatically generated by [workflow](https://github.com/syrupy-project/syrupy/actions?query=workflow%3ACI) using [github-action-benchmark](https://github.com/marketplace/actions/continuous-benchmark).

src/syrupy/session.py
noahnu: If the benchmarking is accurate, there appears to be a bit of a slowdown: https://github.com/syrupy-project/syrupy/pull/927#issuecomment-2563177603

I ran `inv benchmark` locally, and this branch:

```
------------------------------------------------------------------------------ benchmark: 3 tests ------------------------------------------------------------------------------
Name (time in s)         Min               Max              Mean            StdDev            Median               IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_1000x_reads      1.0756 (1.0)      1.2278 (1.0)      1.1648 (1.0)      0.0630 (4.10)     1.1974 (1.0)      0.0948 (3.49)          2;0  0.8585 (1.0)           5           1
test_standard         1.1904 (1.11)     1.3294 (1.08)     1.2822 (1.10)     0.0534 (3.48)     1.2960 (1.08)     0.0439 (1.62)          1;1  0.7799 (0.91)          5           1
test_1000x_writes     1.3974 (1.30)     1.4278 (1.16)     1.4089 (1.21)     0.0153 (1.0)      1.3986 (1.17)     0.0272 (1.0)           1;0  0.7098 (0.83)          5           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

and "main":

```
------------------------------------------------------------------------------ benchmark: 3 tests ------------------------------------------------------------------------------
Name (time in s)         Min               Max              Mean            StdDev            Median               IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_1000x_reads      1.0264 (1.0)      1.1148 (1.0)      1.0855 (1.0)      0.0362 (1.0)      1.1006 (1.0)      0.0474 (1.0)           1;0  0.9212 (1.0)           5           1
test_standard         1.0904 (1.06)     1.1974 (1.07)     1.1461 (1.06)     0.0477 (1.32)     1.1414 (1.04)     0.0876 (1.85)          2;0  0.8725 (0.95)          5           1
test_1000x_writes     1.0937 (1.07)     1.2011 (1.08)     1.1406 (1.05)     0.0405 (1.12)     1.1313 (1.03)     0.0534 (1.13)          2;0  0.8767 (0.95)          5           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

noahnu: > this scales per snapshot location,

This comes up when users are leveraging test parameterization over data sets. i.e. the benchmark case of parameterizing over 1000 test cases isn't too unrealistic from how I've seen syrupy used in the past.

noahnu: FYI, the benchmarking is broken in CI for contributors (something to do with permissions in the github workflow I need to fix), however you should be able to run it locally with `inv benchmark`.

huonw:
Thanks for the review. I'll get back to it when I'm back at work after this holiday period.

huonw:
I've updated this to improve performance and address that concern.

This required making `PyTestLocation` hashable to be able to use it in a dictionary. This gives O(1) access to the relevant snapshot no matter how many snapshot writes are queued, rather than having to do a linear scan of a whole file's queue. The most reliable way to do this is setting `@dataclass(frozen=True)`, but this requires some "hacks" using `object.__setattr__` for initialising the no-init fields.

I've done this work across many independent commits.

Benchmarks on my machine, reporting the mean time &pm; std-dev:

| benchmark | before (`main`) | initial version of this PR | current version |
|---|---|---|---|
| test_1000x_reads | 637.6 &pm; 24.8 | 664.9 &pm; 20.2 | 656.9 &pm; 30.1 |
| test_1000x_writes | 655.2 &pm; 33.4 | 810.1 &pm; 37.6 ‼️ | 675.5 &pm; 53.5 |
| test_standard | 691.8 &pm; 32.5 | 714.9 &pm; 32.5 | 695.0 &pm; 20.4 | 

In summary:

- I can reproduce the large slow-down that we were concerned about in `test_1000x_writes` in the initial version of this PR ✅ 
- the new version seems to get the numbers back to `main` (at least, seems likely to be within normal variation, based on the standard deviation there, and also when I run the benchmarks again locally)

<details><summary>Expand for the full details</summary>

Before (`main`):

```
----------------------------------------------------------------------------------- benchmark: 3 tests -----------------------------------------------------------------------------------
Name (time in ms)          Min                 Max                Mean             StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_1000x_reads      609.2453 (1.0)      665.1862 (1.0)      637.6388 (1.0)      24.7847 (1.0)      642.8187 (1.0)      45.0147 (1.37)          2;0  1.5683 (1.0)           5           1
test_1000x_writes     624.3305 (1.02)     705.0878 (1.06)     655.2347 (1.03)     33.3592 (1.35)     647.2840 (1.01)     51.7828 (1.57)          1;0  1.5262 (0.97)          5           1
test_standard         644.7847 (1.06)     735.3719 (1.11)     691.7773 (1.08)     32.4726 (1.31)     695.5663 (1.08)     32.9661 (1.0)           2;0  1.4456 (0.92)          5           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Initial version of this PR:

```
----------------------------------------------------------------------------------- benchmark: 3 tests -----------------------------------------------------------------------------------
Name (time in ms)          Min                 Max                Mean             StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_1000x_reads      641.2010 (1.0)      688.5666 (1.0)      664.8896 (1.0)      20.2277 (1.0)      657.1012 (1.0)      33.6769 (1.11)          2;0  1.5040 (1.0)           5           1
test_1000x_writes     775.0379 (1.21)     871.4801 (1.27)     810.0917 (1.22)     37.6060 (1.86)     797.2370 (1.21)     45.4147 (1.50)          1;0  1.2344 (0.82)          5           1
test_standard         669.9822 (1.04)     734.9884 (1.07)     714.8819 (1.08)     26.4767 (1.31)     721.9564 (1.10)     30.3537 (1.0)           1;0  1.3988 (0.93)          5           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Current version of this PR:

```
----------------------------------------------------------------------------------- benchmark: 3 tests -----------------------------------------------------------------------------------
Name (time in ms)          Min                 Max                Mean             StdDev              Median                IQR            Outliers     OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_1000x_reads      620.9546 (1.0)      694.4665 (1.0)      656.9467 (1.0)      30.0991 (1.47)     660.8149 (1.0)      50.0408 (1.57)          2;0  1.5222 (1.0)           5           1
test_1000x_writes     624.0394 (1.00)     762.7209 (1.10)     675.4554 (1.03)     53.5257 (2.62)     666.0885 (1.01)     63.9635 (2.01)          1;0  1.4805 (0.97)          5           1
test_standard         668.8180 (1.08)     720.2504 (1.04)     695.0192 (1.06)     20.4068 (1.0)      699.0625 (1.06)     31.8778 (1.0)           2;0  1.4388 (0.95)          5           1
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

</summary>

src/syrupy/session.py
huonw: Updated. See https://github.com/syrupy-project/syrupy/pull/927#issuecomment-2586208297

Thanks for flagging!

src/syrupy/location.py
```diff
@@ -23,27 +23,42 @@ class PyTestLocation:
...
         test_relpath = Path(test_relfile)
-        self.modulename = ".".join([*test_relpath.parent.parts, test_relpath.stem])
-        self.nodename = test_node.replace(f"{self.modulename}.", "")
-        self.testname = self.nodename or self.methodname
+        object.__setattr__(
+            self,
+            "modulename",
+            ".".join([*test_relpath.parent.parts, test_relpath.stem]),
+        )
+        object.__setattr__(self, "methodname", None)
```
huonw, line 57: Setting this attribute is newly added: it previously wasn't being set on this codepath, and attempting to hash/compare the location values (to put into the queue dictionary) was blowing up when accessing it. 

It seems like it was previously not read at all for doc tests?

noahnu (approved):

huonw:
Thank you!

PR #956, opened 2025-02-20
Avoid unnecessary env updates to reduce chances of segfault

sparrowt:
## Description

This PR reduces unnecessary calls to `os.environ.update` (via `env_context`) in order to reduce the probability of a segmentation fault caused by a race with `getenv` in another thread as described by https://github.com/syrupy-project/syrupy/issues/955

## Related Issues

- The single related issue I'm aware of is https://github.com/syrupy-project/syrupy/issues/955 and:
  - this PR should significantly mitigate that issue
  - however there are still cases in which `os.environ` is modified (when `assert` compares 2 `SnapshotAssertion` instances and `--snapshot-no-colors` is enabled) so I'm not sure this PR should be marked as one taht "fixes" that issue
  - a more full solution may be to avoid setting the env at all if it's possible to replace this with use of `contextvars` instead?

## Checklist

- [x] This PR has sufficient documentation -- I think so; it is a bugfix with sufficient explanation in the commit messages
- [ ] This PR has sufficient test coverage -- not yet, I can add this if desired though it will be quite contrived I think

## Additional Comments

Applying this patch locally allowed our playwright test run to complete 200 full runs with zero segfaults (whereas previously it was segfaulting roughly once on every 5-15 runs).

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -1,4 +1,5 @@
 import argparse
+import contextlib
 import sys
 from functools import lru_cache
 from gettext import gettext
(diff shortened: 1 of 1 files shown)


noahnu:
> a more full solution may be to avoid setting the env at all if it's possible to replace this with use of contextvars instead?

Happy to explore that as an option -- will add that to the v5 backlog

noahnu:
@all-contributors add @sparrowt for bug fix

sparrowt:
Thanks so much for the quick response & release as well - much appreciated!

Looks like I can remove my monkey patch and use an official version after all which is always nicer 🎉

PR #978, opened 2025-04-22
fix: Block terminal summary for xdist workers.

paul-ollis:
## Description

This PR prevents pytest_terminal_summary from executing in pytest-xdist worker processes.

I have seen this cause workers to use very high CPU (100% of a core or close) and cause a significant delay at the end of test runs (about 5 seconds).

I cannot see any reason why xdist worker should need to run the terminal summary, so simply preventing execution seems to be a sensible solution. Of course I may be wrong about this.

All tests appear to still pass. It is not obvious to me how to add a test for this change, but I do not think it is really necessary.

## Related Issues

None.

## Checklist

- [ n/a] This PR has sufficient documentation.
- [ x] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary).

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -26,6 +26,7 @@ from .terminal import (
 from .utils import (
     env_context,
     import_module_member,
+    is_xdist_worker,
 )
 
 # Global to have access to the session in `pytest_runtest_logfinish` hook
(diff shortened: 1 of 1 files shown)


noahnu (approved):
The change makes sense to me. The controller prints the report.

noahnu:
@all-contributors add @paul-ollis for bug

noahnu:
The main branch has unreleased changes from v5 on it so I'll have to create a v4 backport branch to get your change out earlier. I'll take a look later this week.

PR #1042, opened 2025-11-29
feat: allow empty lines in snapshot

### LINKED ISSUE #619: .ambr snapshot files have meaningful trailing whitespace
**Describe the bug**

When generating snapshots from mutliline strings the text is idented in the snapshot file. This means that empty lines in multiline strings generate lines containing only (trailling) whitespace in the line.

Trimming this whitespace then invalidates the snapshot causing the test to fail.

**To reproduce**

```
def test(snapshot):
    lines = "\n".join([
        "line 1",
        "", # empty string
        "line 3"
    ])
    assert lines == snapshot
```

This test passes fine.

Then trim trailing whitespace on the generated snapshot.

Now the test fails.

**Expected behavior**

Snapshot files shouldn't have meaningful trailing whitespace.

George-Ogden:
## Description

This PR allows empty lines in the snapshot file to be treated that way, without affecting other snapshot semantics.

## Example
Say I have a test output that looks like this:
```markdown
# Outputs README-like text

Including empty newlines above and below

<footer>Followed by a footer</footer>
```

This is saved in a snapshot as:
```python
  """
  # Outputs README-like text
  
  Including empty newlines above and below
  
  <footer>Followed by a footer</footer>
  """
```
Therefore, there are two lines in this snapshot that include just spaces: `"  "`.
This PR does not affect how the snapshots are saved, but if you trim the empty spaces on the blank lines, such as with a linting tool, the snapshot will load them as empty lines (currently, they are skipped).

## Related Issues

- closes #619, which asks for this feature

## Checklist

## Additional Comments

No additional comments.

### DIFF
--- a/src/syrupy/extensions/amber/serializer.py
+++ b/src/syrupy/extensions/amber/serializer.py
@@ -1,5 +1,6 @@
 import collections
 import inspect
+import textwrap
 from collections import OrderedDict
 from collections.abc import Callable, Generator, Iterable
 from types import (
@@ -184,6 +185,9 @@ class AmberDataSerializer:
                             snapshot_data = ""
                     elif test_name is not None and line.startswith(cls._indent):
                         snapshot_data += line[indent_len:]
+                    elif test_name is not None and line.strip() == "":
+                        # Remove trailing whitespace (but not newline character) from empty line.
+                        snapshot_data += textwrap.dedent(line)
         except FileNotFoundError:
             pass
         else:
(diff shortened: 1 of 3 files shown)


noahnu (changes requested):

noahnu:
> I also think this will affect a very small number of users, and it comes at such a small cost.

I appreciate the proposal/contribution, however this does violate some of the core design principles for syrupy so will reject this particular change. That being said, you should be able to achieve a whitespace insensitive serializer with a custom syrupy extension.

PR #1126, opened 2026-06-26
feat: add --snapshot-no-cleanup to keep unused snapshots on update

### LINKED ISSUE #963: Can we add a flag to just skip deletion of snapshots completely when doing --snapshot-update
I have a use case where I want to run syrupy/pytest with an argument, on the same test suite. It's convenient and I can't actually run all as part of the same test suite by generating tests anyways due to resource constraints (gen ai use case). All I want, is a way to update the snapshots, without deleting the irrelevant ones, and I'm happy to opt out of any/all deletion behaviour.

 a simple flag `--skip-cleanup` or something would be amazing, and solve all my problems

Yes, in theory we could do a bunch of esoteric pytest stuff, and try to enumerate everything, but for this use case it greatly overcomplicates. Plus, we only want to run the snapshots for one given model at a time in practice anyways, and reuse the same test suite vs. having a ton of dupes

**Additional context**

frenck:
## Description

When running `--snapshot-update` against a subset of the test suite (for example filtered with `-k`/`-m`, or when resource constraints mean you cannot run every test in a single invocation), syrupy deletes the snapshots whose tests did not run. There is currently no way to update snapshots without that cleanup.

This adds `--snapshot-no-cleanup`. With it, `--snapshot-update` still writes new snapshots and updates existing ones, but unused snapshots are left in place instead of being deleted, and the run does not fail on them. They are reported as unused rather than deleted.

The behaviour is centralised in a single `should_delete_unused_snapshots` property on the report (`update_snapshots and not no_cleanup`), which both the session deletion gate and the report messaging consult, so the "deleted" vs "unused" wording and the deletion itself stay in sync. Without the flag, behaviour is unchanged.

## Related Issues

- Closes #963, a way to update snapshots without deleting unused ones.

## Checklist

## Additional Comments

Added an integration test that runs `--snapshot-update --snapshot-no-cleanup` against a filtered subset and asserts the unused snapshot is reported as unused and kept on disk rather than deleted. The README CLI options table is updated. Happy to rename the flag (for example `--snapshot-update-no-cleanup`) if you would prefer to scope it explicitly to updates.


noahnu (approved):
Thank you for all these PRs, they're really appreciated!

frenck:
@noahnu You're welcome 👍 

../Frenck  

<p>
  <a href="https://www.linkedin.com/in/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://youtube.com/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://bsky.app/profile/frenck.social"></a>
  &nbsp;&nbsp;
  <a href="https://fosstodon.org/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.instagram.com/frenck/"></a>
  &nbsp;&nbsp;
  <a href="https://www.threads.net/@frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.facebook.com/frenck.dev/"></a>
  &nbsp;&nbsp;
  <a href="https://x.com/frenck"></a>
  &nbsp;&nbsp;
  <a href="https://www.tiktok.com/@frenck.nl"></a>
</p>

Blogging my personal ramblings at <a href="https://frenck.dev">frenck.dev</a>

PR #901, opened 2024-10-15
feat: detect unused snapshots despite xdist

epenet:
## Description

Improve xdist compatibility - with the ability to serialize/deserialize the report data.
See https://github.com/home-assistant/core/pull/128162 for sample implementation

## Related Issues

> Known Limitations
> pytest-xdist support only partially exists. There is no issue when it comes to reads however when you attempt to run pytest --snapshot-update, if running with more than 1 process, the ability to detect unused snapshots is disabled. See https://github.com/syrupy-project/syrupy/issues/535 for more information.
> We welcome contributions to patch these known limitations.

## Checklist

- [ ] This PR has sufficient documentation.
- [ ] This PR has sufficient test coverage.
- [x] This PR title satisfies semantic [convention](https://www.conventionalcommits.org/en/v1.0.0/#summary).

## Additional Comments

No additional comments.

PR #1132, opened 2026-06-26
feat: detect unused snapshots under pytest-xdist

### LINKED ISSUE #535: Incompatible with pytest-xdist / thread safety

frenck:
## Description

Running with `pytest-xdist` (`-n`) disabled unused snapshot detection: each worker only sees the tests it ran, so it can't tell whether a snapshot is unused or just owned by a test that ran on another worker. As a result `--snapshot-update` under xdist never removed obsolete snapshots, and plain runs never failed on them.

This makes it work. Each worker publishes the snapshots it touched on `config.workeroutput`; the controller harvests them in `pytest_testnodedown` and combines them, then runs the existing unused-detection and removal path. Detection, the failure exit status, and partial in-file removal all behave the same as a single-process run.

This builds on @epenet's earlier attempt in #901, which was reverted because serializing the full report to disk on every run was too slow on large suites. The two changes that address that here:

- Transport is `config.workeroutput` (delivered over xdist's existing channel), not temp files and `json.dump`.
- The collected-item list, which is identical on every worker and is the large payload, is sent from a single worker (`gw0`) instead of once per worker.

The controller rebuilds just enough to reuse the normal logic: the snapshot collections, the per-location extension (so partial removal within a shared file works), the xfail count, and a stand-in for each collected item so `PyTestLocation` can resolve names.

## Related Issues

- Closes #535, unused snapshot detection disabled under pytest-xdist.
- Supersedes #901 by @epenet, same goal, with the performance issue that caused its revert addressed.

## Checklist

## Additional Comments

New `tests/integration/test_snapshot_xdist.py` covers detection and partial removal under `-n 2`. Many integration tests previously marked `plugin_args_fails_xdist` now pass under xdist and were switched to `plugin_args`; three that fail for unrelated reasons (failure-diff output, `--pyargs` node selection, custom-comparator approximate match) are kept on the xfail fixture.

Two things worth a maintainer's eye:

- This is a behavior change: an xdist run with unused snapshots now fails (or deletes them under `--snapshot-update`), matching non-xdist behavior. Surfaced as a true positive, but it can turn a previously-green xdist suite red on upgrade.
- The collected-item list is sent from `gw0` only, which assumes default load scheduling. Reconstructed items also don't carry doctest state, so snapshotting inside doctests under xdist isn't covered.

@epenet, since this revives your #901 and the motivation was Home Assistant, could you sanity-check the performance side and whether the reporting semantics match what you needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ See [#675](https://github.com/syrupy-project/syrupy/issues/675) for the original
 
 ## Known Limitations
 
-- `pytest-xdist` support only partially exists. There is no issue when it comes to reads however when you attempt to run `pytest --snapshot-update`, if running with more than 1 process, the ability to detect unused snapshots is disabled. See [#535](https://github.com/syrupy-project/syrupy/issues/535) for more information.
+- `pytest-xdist`: unused snapshot detection and removal are supported when running with multiple processes. Each worker reports the snapshots it used and the controller combines them to determine which snapshots are unused. See [#535](https://github.com/syrupy-project/syrupy/issues/535) for more information.
 
 _We welcome contributions to patch these known limitations._
 
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -210,6 +210,20 @@ def pytest_collection_finish(session: Any) -> None:
     session.config._syrupy.select_items(session.items)
 
 
+def pytest_testnodedown(node: Any, error: Any) -> None:
+    """
+    Collect a pytest-xdist worker's snapshot report as the worker shuts down.
+    Runs on the controller; the payload is published via ``config.workeroutput``.
+    https://pytest-xdist.readthedocs.io/en/stable/distribution.html
+    """
+    workeroutput = getattr(node, "workeroutput", None)
+    syrupy = getattr(node.config, "_syrupy", None)
+    if workeroutput and syrupy:
+        report = workeroutput.get("syrupy_report")
+        if report is not None:
+            syrupy.add_worker_report(report)
+
+
 def pytest_runtest_logreport(report: pytest.TestReport) -> None:
     """
     After each of the setup, call and teardown runtest phases of an item.
--- a/src/syrupy/data.py
+++ b/src/syrupy/data.py
@@ -122,6 +122,21 @@ class SnapshotCollections:
     def __contains__(self, key: str) -> bool:
         return key in self._snapshot_collections
 
+    def serialize(self) -> dict[str, list[str]]:
+        """Reduce to a ``{location: [snapshot name, ...]}`` mapping for transport."""
+        return {
+            location: [snapshot.name for snapshot in collection]
+            for location, collection in self._snapshot_collections.items()
+        }
+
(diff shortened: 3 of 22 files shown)


epenet:
TBH my previous attempt was over 18 months ago and I'm not sure I have the right mindset to go digging into pytest/syrupy once more

README.md
```diff
@@ -575,7 +575,7 @@ See [#675](https://github.com/syrupy-project/syrupy/issues/675) for the original
 
 ## Known Limitations
 
-- `pytest-xdist` support only partially exists. There is no issue when it comes to reads however when you attempt to run `pytest --snapshot-update`, if running with more than 1 process, the ability to detect unused snapshots is disabled. See [#535](https://github.com/syrupy-project/syrupy/issues/535) for more information.
+- `pytest-xdist`: unused snapshot detection and removal are supported when running with multiple processes. Each worker reports the snapshots it used and the controller combines them to determine which snapshots are unused. See [#535](https://github.com/syrupy-project/syrupy/issues/535) for more information.
```
noahnu, line 578: Not really a "limitation" at this point. You mentioned a limitation around default load scheduling? that's what I'd call out. Will get this merged as is though

noahnu (approved):
Changes LGTM -- I acknowledge the behavioural change, though pytest-xdist compatibility was already broken before so I think it's fine with a warning in the changelog without marking as a breaking change.

edgarrmondragon:
This seems to have introduced a bug where tests fail with a `pluggy._manager.PluginValidationError` when pytest-xdist is not installed alongside syrupy, so I went ahead and opened a PR:

- https://github.com/syrupy-project/syrupy/pull/1140/

PR #1162, opened 2026-07-18
fix(matchers): replace values at nested paths in path_value default mode

chuenchen309:
### Problem

`path_value` in its default mode (`regex=False`) never replaces a value whose path key contains a regex-special character — which is **every nested path**, since the `.` path separator is special:

```python
matcher = path_value({"user.token": "abc12345"}, types=(str,), replacer=redact)
# {"user": {"token": "abc12345"}} serializes with "abc12345" left in place;
# the identical matcher with regex=True redacts it.
```

`path_type` matches the path with `_path_match`, which `re.escape`s the key when `regex=False`, so the `re.Match` it hands to the replacer carries the *escaped* pattern (`user\.token`). The replacer then looks that escaped pattern up against the original, unescaped mapping keys (`user.token`), misses, and returns the value unchanged. For the redaction use case this feature exists for, the value silently leaks into the committed snapshot.

Top-level keys (no special char) and `regex=True` are unaffected, which is why the existing tests — all `regex=True` — didn't catch it.

### Fix

Key the value lookup by the same escaped form the path match used, so the lookup agrees with the pattern `path_type` produces.

### Tests

`test_path_value_replaces_nested_value_in_literal_mode` serializes a nested value with a default-mode `path_value` matcher and asserts it is replaced and does not leak; it fails before this change and passes after. The rest of `tests/syrupy/extensions/amber/` stays green (91 passed). `ruff check`, `ruff format`, and `mypy --strict src` pass.

---

*Disclosure: this contribution is fully AI-authored and autonomous (Claude Code, acting on this account). An AI found the bug, wrote and ran the repro and the test, and wrote this description; the human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.*

### DIFF
--- a/src/syrupy/matchers.py
+++ b/src/syrupy/matchers.py
@@ -90,15 +90,20 @@ def path_value(
     Matches the path regex against the string repr of values for the types specified
     """
 
+    regex = kwargs.get("regex", False)
+    # path_type matches the path with _path_match, which escapes the key when
+    # regex is off, so the Match handed back carries the escaped pattern. Key
+    # the value lookup by that same form, or a path key with a regex-special
+    # char (e.g. the "." joining a nested path) never matches and the value is
+    # silently left unreplaced.
+    value_patterns = {(k if regex else re.escape(k)): v for k, v in mapping.items()}
     kwargs["mapping"] = dict.fromkeys(mapping, types)
     kwargs["replacer"] = lambda data, path_matches: (
         replacer(
             data,
-            _path_match(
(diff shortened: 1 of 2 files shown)


noahnu:
Thank you for the contribution! The change makes sense to me, though note that it is a breaking change and will go out with Syrupy v6 (no ETA).

noahnu (approved):

noahnu:
@all-contributors add @chuenchen309 for bug

PR #1175, opened 2026-07-25
feat: support in-memory snapshot diff data

### LINKED ISSUE #608: Store snapshot diff without storing previous snapshot

Current API requires you you to store a previous snapshot before the next "diff" can be stored.

For example

```python
    assert result == snapshot
    assert result_after_something_happens == snapshot(diff=0)
``` 

If my test just depends on the diff snapshot (checking that the changes always remain the same) it's not possible to only snapshot the diff. 

I'd be great if I could do something like this:
```python
    assert result_after_something_happens == snapshot(diff=result)
```

yangfan-yf-yf:
## Description

Allow `snapshot(diff=...)` to use an in-memory data object as the diff baseline. Non-integer, non-string values are serialized with the assertion's existing options and used to generate the diff snapshot, so a separate base snapshot is not written.

Integer and string `diff` values retain their existing behavior as snapshot indexes and names.

## Related Issues

- Closes #608, support storing only a snapshot diff when the base data is already available in the test.

## Checklist

## Additional Comments

- Added a regression test and Amber snapshot for an in-memory dictionary baseline.
- Verified with `py -m uv run inv test` (`473 passed, 5 xfailed, 6 xpassed`) and `py -m uv run inv lint`.


src/syrupy/assertion.py
```diff
@@ -270,7 +270,7 @@ def __with_prop(self, prop_name: str, prop_value: Any) -> None:
     def __call__(
         self,
         *,
-        diff: Optional["SnapshotIndex"] = None,
+        diff: Optional["SnapshotIndex | SerializableData"] = None,
```
noahnu: Might make more sense to have a new kwarg entirely for this. That way you can also diff on a string, e.g. diff_data= or something like that.

yangfan-yf-yf:
Updated in `53c9e0b` and rebased on current `main`.

`diff` is again limited to a saved snapshot index or name. The new `diff_data` keyword holds the in-memory baseline, so strings are treated as data rather than being ambiguous with snapshot names. The README now documents both paths, and the Amber regression coverage includes dictionary and string baselines.

Validation:

```text
.venv\Scripts\python.exe -m pytest tests/syrupy/extensions/amber/test_amber_snapshot_diff.py -q
4 passed, 1 xfailed

.venv\Scripts\mypy.exe src
Success: no issues found in 25 source files

.venv\Scripts\ruff.exe check .
All checks passed

.venv\Scripts\ruff.exe format --check .
100 files already formatted
```

noahnu (approved):

noahnu:
@all-contributors add @yangfan-yf-yf for feature

noahnu:
@all-contributors add @yangfan-yf-yf for code

PR #1196, opened 2026-08-09
feat: mark tests that use snapshots

### LINKED ISSUE #922: CLI option to update snapshots without running other tests

I am using syrupy in a few libraries, and I find myself in a common pattern in which I write a Makefile target like `make snapshot-update` which invokes `pytest --snapshot-update`.  But the real intention behind that make target is not necessarily to run all the tests _and_ update snapshots; but to update the snapshots only.

I think it'd be a neat feature to support a pytest selector which isolates only the tests depending on the snapshot fixture; so that `pytest --snapshot-update` could be extended to only run tests with any snapshots.

It's not a big deal! I just run all my tests when i run updates. :smile: 

**Additional context**

yangfan-yf-yf:
## Description

Mark every test whose fixture closure includes Syrupy's `snapshot` fixture with `syrupy_snapshot`.

Users can update snapshots without running unrelated tests with:

```shell
pytest --snapshot-update -m syrupy_snapshot
```

- rely on pytest's built-in marker selection instead of maintaining separate selection logic
- cover direct and indirect snapshot fixture dependencies
- register the marker for strict marker validation
- preserve existing `--snapshot-update` behavior and snapshot cleanup bookkeeping when `-m` or `-k` deselects tests

## Related Issues

- Closes #922.

## Checklist

- [x] This PR title satisfies semantic convention.

## Additional Comments

Validation:

- marker integration tests across serial, xdist disabled, and two xdist workers: 12 passed
- related update and unused-snapshot regressions: 64 passed, 2 xfailed
- full test suite: 487 passed, 5 xfailed, 6 xpassed
- Ruff check and format check passed for the changed Python files
- strict Mypy passed for `src` and `benchmarks`
- `git diff --check` passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,14 @@ pytest --snapshot-update
 
 A snapshot file should be generated under a `__snapshots__` directory in the same directory as `test_file.py`. The `__snapshots__` directory and all its children should be committed along with your test code.
 
+Syrupy automatically marks tests that directly or indirectly use the `snapshot`
+fixture with `syrupy_snapshot`. To update snapshots without running unrelated
+tests, combine the marker with the update option:
+
+```shell
+pytest --snapshot-update -m syrupy_snapshot
+```
+
 #### Usage in `unittest.TestCase` subclasses
 
 [Due to limitations in `unittest` and `pytest`](https://docs.pytest.org/en/9.0.x/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses), the `snapshot` fixture is not directly usable in `TestCase` subclasses (including Django's `TestCase`).
--- a/src/syrupy/__init__.py
+++ b/src/syrupy/__init__.py
@@ -192,6 +192,7 @@ def pytest_sessionstart(session: Any) -> None:
     session.config._syrupy.start()
 
 
+@pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(
     session: Any, config: Any, items: list["pytest.Item"]
 ) -> None:
@@ -199,6 +200,9 @@ def pytest_collection_modifyitems(
     After tests are collected and before any modification is performed.
     https://docs.pytest.org/en/latest/reference.html#_pytest.hookspec.pytest_collection_modifyitems
     """
+    for item in items:
+        if "snapshot" in getattr(item, "fixturenames", ()):
+            item.add_marker("syrupy_snapshot")
     config._syrupy.collect_items(items)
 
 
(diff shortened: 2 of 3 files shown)


noahnu:
Pytest supports filtering tests by marks. What if we auto-marked all tests with some "syrupy-snapshot" mark? Then users can use the built-in pytest filter behaviour? It'd keep the complexity that we have to maintain on the syrupy side down

yangfan-yf-yf:
Good suggestion. I replaced the custom option and selection logic with a `syrupy_snapshot` marker applied by the plugin. Tests that directly or indirectly use the fixture receive it before pytest's marker filtering runs, so `pytest --snapshot-update -m syrupy_snapshot` covers the use case with much less plugin-side logic.

I also added coverage for direct and indirect fixtures, composition with `-k` and existing marks, unused-snapshot cleanup, xdist, and unchanged unfiltered updates. The full suite passes on the updated branch (`487 passed, 5 xfailed, 6 xpassed`).

noahnu:
Let's test indirect cases as well:

```py
import pytest

from syrupy.extensions.json import JSONSnapshotExtension


@pytest.fixture
def snapshot_json(snapshot):
    return snapshot.with_defaults(extension_class=JSONSnapshotExtension)
    # or return snapshot.use_extension(JSONSnapshotExtension)


def test_api_call(client, snapshot_json):
    resp = client.post("/endpoint")
    assert resp.status_code == 200
    assert snapshot_json == resp.json()
```

Since in this case the fixture is snapshot_json right? Do we need to walk the fixtures? Or does it get resolved to the final fixtures by the time we check the fixture names?

I'll take a closer look later this week as well. Thank you for the contribution!

yangfan-yf-yf:
Yes. By the time `pytest_collection_modifyitems` runs, `item.fixturenames` contains the resolved fixture closure, not only the fixture names written in the test signature. For a test requesting `snapshot_json`, the list also contains its transitive `snapshot` dependency, so the marker is applied.

I updated the regression in `bd1eeab` to use the exact extension pattern from your example:

```python
@pytest.fixture
def snapshot_json(snapshot):
    return snapshot.with_defaults(extension_class=JSONSnapshotExtension)
```

The test now verifies that both the direct `.ambr` snapshot and the indirect JSON-extension snapshot are written, and it runs in serial, `-n 0`, and `-n 2` modes. The focused marker file passes (`12 passed`), the complete suite passes (`487 passed, 4 xfailed, 7 xpassed`), and Ruff, format, and strict Mypy checks pass.

noahnu (approved):

PR #1207, opened 2026-08-17
perf: narrow unused snapshot matching to plausible test files

### LINKED ISSUE #1193: Performance regression in syrupy 5.5.0 with pytest-xdist + pytest-split
I currently run tests with both pytest-xdist and pytest-split combined and noticed a significant performance regression with 5.5.0. 

Environment
syrupy 5.5.3, pytest 9.1.1, pytest-xdist 3.8.0, pytest-split 0.11.0, Python 3.10
~1,460 snapshot files / ~3,760 snapshots; ~8.8k collected items, ~2.2k selected per shard.

Same machine, same shard, same 2,205 tests, invocation pytest -n auto --splits 4 --group 1:

Version | Wall
-- | --
5.5.3 + --snapshot-warn-unused | 324s
5.4.0 | 70s

Checking the code a bit, it does look like the issue is that there is a difference in how pytest-split interacts with syrupy collection:
collected items = whole suite (captured in pytest_collection_modifyitems, pre-deselection)
selected items = this shard's quarter (captured in pytest_collection_finish)

Which causes an explosion of `_ran_items_match_name` to be run due to the unused snapshots. This setup can't really be accurate, as the other groups run in separate processes, so the controller merges its xdist workers and still only sees a quarter of the suite. Those snapshots are reported unused on every shard regardless.

An additional issue is that `--snapshot-warn-unused` doesn't avoid the work — it only changes the styling and exit code, `num_unused` is still computed.

Would it be reasonable to skip unused detection when the selection is partial?

tiagoskaneta:
## Description

Unused snapshot detection matched every discovered snapshot against every test that ran. When the whole collected suite is selected that work is skipped, but a sharding plugin such as pytest-split selects a fraction of the collected items, so `selected_all_collected_items` is False and the per-snapshot path is taken for the entire snapshot corpus. The cost grew with `discovered snapshots x tests that ran`, and each comparison rebuilt a `PyTestLocation` and called `Path.resolve()`, so it was dominated by filesystem calls. Suites storing one file per snapshot feel this most, because every snapshot is then its own collection location.

Measured on a suite of 1600 single-file snapshots with a quarter of the items selected, unused detection goes from 10.98s to 1.04s. On a real 2200-test shard it takes the run from 361s back to 78s, against 81s on the 5.4.0 release that predates cross-worker detection.

## Related Issues

Closes #1193 

## Checklist

## Additional Comments

AI used for analysis and fix.

### DIFF
--- a/src/syrupy/report.py
+++ b/src/syrupy/report.py
@@ -189,7 +189,7 @@ class SnapshotReport:
     def num_updated(self) -> int:
         return self._count_snapshots(self.updated)
 
-    @property
+    @cached_property
     def num_unused(self) -> int:
         return self._count_snapshots(self.unused)
 
@@ -213,7 +213,7 @@ class SnapshotReport:
             if self.selected_items[nodeid]
         )
 
-    @property
+    @cached_property
     def unused(self) -> "SnapshotCollections":
         """
         Iterate over each snapshot that was discovered but never used and compute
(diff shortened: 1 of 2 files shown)


noahnu:
Thank you for the contribution! I'll take a closer look later this week

noahnu (approved):

noahnu:
@all-contributors add @tiagoskaneta for code

PR #1239, opened 2026-09-12
fix: opt-in file lock for concurrent amber writes under xdist

### LINKED ISSUE #1237: `--snapshot-update` under `pytest-xdist` silently destroys most of a multi-entry `.ambr` file
## `pytest-xdist` workers race on writing .ambr files, silently destroying snapshots

When `--snapshot-update` runs under `pytest-xdist` and more than one worker has updates for the same amber file, `AmberDataSerializer.write_file` performs an unsynchronized read-modify-write on that file. Workers read the file while another worker is streaming it out, get a truncated parse, merge their own snapshots into that partial content, and write it back. Snapshots are lost.

pytest exits 0 and the summary reports every snapshot as successfully updated, so nothing signals the loss. It surfaces later as a "_deleted snapshot_" in the diff, which reads like an unused-snapshot removal but is not one.

This is not the unused-snapshot detection added in #1132. In the failing runs `SnapshotSession.remove_unused_snapshots` is never called, the report's `unused` set is empty. The data is lost in the writes.

## To reproduce

You'll need this file:

`test_race.py`
```python
import pytest

VALUE = "v1"  # bump to v2 to make every snapshot need an update

@pytest.mark.parametrize("i", range(200))  # increase this number if needed on your machine
def test_many(snapshot, i):
    assert f"{VALUE}-{i}" == snapshot
```

Then, assuming you have `uv` installed on your machine:
```bash
$ uv init
$ uv add pytest syrupy pytest-xdist
$ uv run pytest test_race.py --snapshot-update -q
200 snapshots generated.
$ grep -c '^# name:' __snapshots__/test_race.ambr  # just to check that .ambr file is consistent
200
$ sed -i 's/VALUE = "v1"/VALUE = "v2"/' test_race.py  # replace v1 by v2
$ uv run pytest test_race.py --snapshot-update -q -n auto  # or -n number_of_workers
200 snapshots updated.                               # reports full success, exits 0
$ grep -c '^# name:' __snapshots__/test_race.ambr
44                                                # 156 snapshots destroyed on my side... 🥲
```

The serial control (same command without `-n auto`) leaves 200/200 every time.

Note that snapshots which merely *passed* are the most vulnerable: a worker only holds the snapshots it rewrote in its own queue, so everything else survives purely through the read-modify-write base. On a large suite where only a handful of snapshots change, a torn base read discards the untouched majority.

## Environment

- syrupy 6.0.0 (latest)
- pytest 9.1.1 (latest)
- pytest-xdist 3.8.0 (latest)
- Python 3.13.9
- Linux 7.0.0, 22 logical CPUs (`-n auto` ➜ 22 workers)

noahnu:
## Summary

Addresses #1237: under `--snapshot-update` with `pytest-xdist`, multiple workers can race on the same multi-entry `.ambr` file. Each worker does an unsynchronized read-modify-write in `AmberDataSerializer.write_file`, so one worker can read a torn or stale base, merge only its own updates, and overwrite peers' work. Pytest still reports a full successful update; the loss shows up later as mysterious "deleted" snapshots.

This is separate from unused-snapshot detection (#1132 / #535). Those paths merge *reports* on the controller. Snapshot *data* was still written independently by each worker.

**Note:** This PR does **not** close #1237 yet. The fix is opt-in (`--snapshot-file-lock`) so default behavior is unchanged until we validate with affected users and make locking the default in a follow-up.

## Approach

Opt-in via `--snapshot-file-lock` (off by default):

1. **Exclusive file lock** around the amber RMW (`fcntl.flock` / `msvcrt` — stdlib, no new deps)
2. **Atomic replace** (write `.tmp`, then replace) so readers never see a half-written file
3. **Configurable lock timeout** via `--snapshot-file-lock-timeout` (default **60s**)
4. **Sidecar cleanup** after the run: workers report written paths; the controller (or single-process session) removes `.lock` / `.tmp` once all writers are done. Lockfiles are not deleted on unlock, which would race with concurrent openers. Locations are tracked before the write so failed/timed-out acquires still get cleaned up.
5. **Warning** when `--snapshot-update` runs under xdist without `--snapshot-file-lock`, noting the experimental flag and that it is expected to become the default in a future minor release.

File locking was chosen over controller-merged writes so same-machine xdist keeps the existing per-worker write path with minimal protocol change. Remote/NFS remains unsupported; controller-side merge is a possible future alternative.

Without the flag, write behavior is unchanged (aside from the warning and always-on compression of the xdist `collected` worker payload to shrink IPC size on large suites).

## Limitations

Same-machine workers only. Remote xdist / NFS still unreliable for advisory locks. Hard kill (`SIGKILL`) can leave a harmless leftover `.lock`.

## Test plan

- [x] Concurrent merge unit test (locked vs unlocked lost-update)
- [x] xdist integration: update with `--snapshot-file-lock` preserves all snapshots
- [x] Warning emitted for xdist + `--snapshot-update` without the flag; suppressed when flag is on
- [x] Lock timeout / Windows lock-file creation / sidecar tracking on write failure


noahnu:
### Review follow-ups in this update

- **Opt-in stays** for now (default still races). Intentionally **not** using `Fixes #1237` so the issue stays open until we default-on after validation — please try `--snapshot-experimental-file-lock` and confirm on #1237.
- **Windows CI:** fixed controller `KeyError` when `collected` was missing after a worker failed to publish (every worker now sends compressed `collected`; `ran_items` skips unknown nodeids). Also hardened Windows lock acquire (timeout) and atomic replace (retries + clearer errors).
- **`delete_snapshots`:** full RMW under the same exclusive lock when the flag is on (no nested lock).
- **Discovery:** only skip real sidecars (`*.ext.lock` / `*.ext.tmp`), not single-suffix names like `photo.tmp`.
- **Tracking:** session-owned write locations only (dropped the process-global sidecar set).

